PR #33, opened 2026-03-24
Fix hashrate chart stuck after navigating to settings and back

pavlenex:
The Hashrate History chart would get stuck after visiting the Settings page and returning to the Dashboard.

According to Claude:
> On remount, the hook immediately recorded a 0 sample (from the loading state) and then only re-sampled when the hashrate value actually changed — so on a stable miner it would freeze on that single zero point for several minutes. Replaced the change-triggered effect with a fixed 5-second interval that skips zero values, so the chart collects data reliably and recovers instantly after navigation.

I tested and it works, but needs code review.

## Before 

https://github.com/user-attachments/assets/b1b06ff9-bc5a-470c-8a24-d9b58277d24e

Stuck like this forever...

## After
https://github.com/user-attachments/assets/915c887e-44b9-49fd-a327-b28bf3a8d1b2

### DIFF
--- a/src/hooks/useHashrateHistory.ts
+++ b/src/hooks/useHashrateHistory.ts
@@ -18,41 +18,34 @@ const SAMPLE_INTERVAL_MS = 5000; // Sample every 5 seconds
  */
 export function useHashrateHistory(currentHashrate: number | undefined): HashrateDataPoint[] {
   const [history, setHistory] = useState<HashrateDataPoint[]>([]);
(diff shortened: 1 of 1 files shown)


GitGab19 (commented):
The problem on the current setup which this PR is not solving yet is that we're not really persisting any historical hashrate data.

To handle the chart properly, we have to start persisting it for sure.

PR #64, opened 2026-04-05
fix: prevent false zero injection in hashrate chart on refresh

### LINKED ISSUE #57: Hashrate chart resets to zero on browser page refresh
When a user refreshes the browser (F5 / Cmd+R), the hashrate chart briefly shows zero before recovering. This does not happen when navigating away to Settings and back.

https://github.com/user-attachments/assets/5e8a371e-a46e-4b68-be23-f48c0ff0802a

rx18-eng:
## Body

when you refresh the page, poolGlobal is undefined while the api is still loading so totalHashrate defaults to 0. the hook immediately samples that zero and saves it to localstorage, which makes the chart dip to zero before recovering.

fix is simple - pass undefined instead of 0 to the history hook until pool data actually loads. the hook already has a guard that skips undefined values so no false zero gets injected.

fixes #57


rx18-eng:
@pavlenex , would appreciate a review whenever you get a moment !

PR #76, opened 2026-04-06
add inline settings editing for pool, identity, and template mode

### LINKED ISSUE #21: Allow settings changes and adjustments without reconfiguring everything from start.
Currently, to change a setting, user is forced to go through the entire reconfiguration process. This is not ideal UX.

We should ensure basic settings such as pool, payout address, tproxy mode, are all editable from the UI, where user saves settings and we restart and save to apply them.

Image

This is something mentioned already during the review of https://github.com/stratum-mining/sv2-ui/pull/17 and the PR here already did it and can serve as an inspiration https://github.com/stratum-mining/sv2-ui/pull/16

rx18-eng:
## summary

so right now if you want to change your pool or username or template mode you have to go through the whole setup wizard again which is annoying. this adds inline editing directly in the settings tab - click the edit icon, change what you need, save, and services restart with the new config.

also pulled out the pool presets into a shared file (src/lib/pools.ts) because both the wizard and settings need the same pool list. custom pool option is there too if you want to enter your own connection details, and save is disabled until address and pubkey are filled in so you can't accidentally save an empty config.

for template mode switching — going from JD to no-JD just clears the jdc/bitcoin stuff, but going the other way needs bitcoin core config so it redirects you to the wizard for that.

## test plan

- go through setup wizard then check settings tab
- try editing pool, identity, template mode
- try custom pool with empty fields (save should be disabled)
- cancel mid-edit should revert everything

closes #21

### DIFF
--- a/src/components/settings/ConfigurationTab.tsx
+++ b/src/components/settings/ConfigurationTab.tsx
@@ -4,9 +4,10 @@ import { useQueryClient } from '@tanstack/react-query';
 import { Card, CardContent, CardDescription, CardHeader, CardTitle } from '@/components/ui/card';
 import { Button } from '@/components/ui/button';
 import { Badge } from '@/components/ui/badge';
-import { InlineEditField } from '@/components/ui/inline-edit-field';
+import { PoolIcon } from '@/components/ui/pool-icon';
 import { useSetupStatus } from '@/hooks/useSetupStatus';
 import { useControlApi, getCurrentConfig } from '@/hooks/useControlApi';
+import { getPoolsForMode, type KnownPool } from '@/lib/pools';
 import type { SetupData } from '@/components/setup/types';
 import {
   Loader2,
@@ -14,6 +15,9 @@ import {
   RotateCw,
   StopCircle,
   Trash2,
+  Pencil,
+  Check,
+  X,
 } from 'lucide-react';
 
 function clearPersistedDashboardState() {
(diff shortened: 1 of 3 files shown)


rx18-eng:
@pavlenex  would appreciate a review whenevr you are free !

pavlenex:
@rx18-eng could you please share a screen recording and test this on your side as well? I’m not able to fully validate every PR on my own, especially for issues that go beyond good first issues. It would really help if authors put some effort into testing and demonstrating their changes too.

rx18-eng:
yeah sure i have made a video sharing it now !

rx18-eng:
@pavlenex lmk if any change is needed ! 
 

https://github.com/user-attachments/assets/e2af5e9c-27cf-44f1-ba61-1d48bc85e62a

rx18-eng:
@pavlenex shall i modify anything else on this pr ?

pavlenex:
@rx18-eng can you rebase and resolve conflicts?

rx18-eng:
@pavlenex i have resolved the conflicts and rebased !

pavlenex:
@rx18-eng Can you ensure consistency in how we handle things? For example here when we change pool name there's no save and restart, we do it in the background, but I think we should tell users restart will occur just like you do in other tabs above, as it may cause operational downtime.

Screenshot 2026-04-17 at 15 23 22

Also once you're editing, you need to click cancel in order to escape to another field to edit, don't think that's optimal.

rx18-eng:
https://github.com/user-attachments/assets/28c94423-b644-40ba-a703-19625af1cae7

rx18-eng:
@pavlenex  how does this look to you ?

pavlenex (approved):
LGTM good work @rx18-eng , there's still polishing here to be done, UI is beginning to lose consistency with a lot of different patterns we keep introducing and also we should allow user to change all options there, including Block Reward Address for example, but that can be done in a follow up PR.

rx18-eng:
yeah sure will look on this and raise a follow up pr soon . thanks for your review !

PR #77, opened 2026-04-07
feat: add DMND as pool option with token-based endpoint discovery

### LINKED ISSUE #73: add DMND as pool option (JD and no-JD)

Anshumancanrock:
Closes #73

### Description

Adds DMND as a selectable pool in the setup wizard (both pool-mining modes, with and without JD).

Unlike other pools that have a static address, DMND uses token-based endpoint discovery. So the flow is: user picks DMND → enters their KYB token → hits "Discover Endpoint" → we resolve the actual pool address via the DMND API and populate the config.

### Changes

 1) New `POST /api/dmnd/discover` route that takes the user's token, calls the DMND discovery API (`/api/pool/urls`), and returns the best endpoint.
 
 2) DMND added to both `POOL_MINING_NO_JD` and `POOL_MINING_JD` lists. Token input + discover button UI shown when DMND is selected. Back-navigation preserves the discovered endpoint.
 
3) Added optional `token` field to `PoolConfig`.

### Screeshorts:

No-JD mode :

Screenshot 2026-04-07 053013

Screenshot 2026-04-07 053117

JD mode:

Screenshot 2026-04-07 053435


Anshumancanrock:
Hi @pavlenex @plebhash , DMND is now available as a selectable pool in both mining modes (with and without JD). Since DMND uses token-based endpoint discovery rather than a static address, the flow differs from other pools. The user provides their KYB token, and we call the DMND discovery API to resolve the actual pool endpoint.

Please review when you have time and let me know if you'd like any changes. Thanks!

pavlenex:
@Anshumancanrock Appreciate the effort, but if you tested this you'd see it doesn't work. This change isn't only using AI to make UI, it needs some backend adjusting and decisions as described in #73

Anshumancanrock:
Hi @pavlenex, Thanks for the feedback. After going through @plebhash's analysis on #73, i understand the issue now. The UI changes alone aren't enough because the SRI Translator and JDC containers don't currently support a device_id config parameter, which is needed to send the DMND KYB token in `SetupConnection.device_id` (format: <random16>::POOLED::<TOKEN>).

My initial approach was to get the UI and endpoint discovery part in place first, with the intention of following up with the backend changes. I should have been upfront about that in the PR description instead of presenting it as a complete solution.

So the actual path would be:

First, add device_id support upstream in sv2-apps (Translator + JDC config parsing).
Then update sv2-ui's config generator to include device_id in the TOML when DMND is selected.
The UI/endpoint discovery parts from this PR can be reused at that point.

I'd like to help with this if you're open to it. Should I start with a PR to sv2-apps for the device_id config support. Also happy to follow whatever direction @plebhash prefers regarding `device_id` vs AllocateMiningJobToken.user_identifier as discussed in #73.

Issue #85, opened 2026-04-08
log-based UI errors


plebhash:
this is meant to be an umbrella tracker for log-based errors

a prerequisite for this is https://github.com/stratum-mining/sv2-ui/issues/82#issuecomment-4207061453

Issue #87, opened 2026-04-08
failed handshake should show specific error


plebhash:
logs usually look like this:

<img width="947" height="614" alt="Image" src="https://github.com/user-attachments/assets/007aa107-8747-49ad-999d-b0b8e5e85da1" />

plebhash:
even though system clock is a common root cause for this kind of error, I don't think we should explicitly mention that on the error card

this should simply inform the user that a handshake failed

then when they look into FAQ (https://github.com/stratum-mining/sv2-ui/issues/67#issuecomment-4193399831) they'll realize that they should check their system clock and try again

Issue #86, opened 2026-04-08
bad braiins username should show specific error


plebhash:
as described in https://github.com/stratum-mining/sv2-ui/issues/82#issuecomment-4207061453

lucasbalieiro:
I'll start working on this one

Issue #83, opened 2026-04-08
unavailable `node.sock` error case is displaying misleading error


plebhash:
@JayrMotta reported that when the Bitcoin Core `node.sock` is not available, the UI is displaying a misleading error

<img width="1932" height="502" alt="Image" src="https://github.com/user-attachments/assets/8bd5c5f0-fd5f-4c34-a7a6-c178d8e4e52f" />

plebhash:
one solution for this could come from the TS modules established on #82

ideally #82 should bootstrap TS code not only to display container logs on UI, but also to parse them so that the UI is able to display accurate information on different scenarios

---

on this specific scenario, whatever error logs come from JDC (see https://github.com/stratum-mining/sv2-apps/issues/418) should help the UI inform the user about the fact that the provided Bitcoin Core `node.sock` is not available

plebhash:
but also, when the user configures the Bitcoin Core `node.sock` for the first time (screenshot below), the TS code should make one initial validation to make sure that's available and prevent the user from going forward in case it's not

of course, after this initial validation, Bitcoin Core could still become unavailable later, so the log parsing described above should still be implemented

<img width="657" height="772" alt="Image" src="https://github.com/user-attachments/assets/22064c6d-5bda-431c-b08e-97807dd8853e" />

plebhash:
with https://github.com/stratum-mining/sv2-apps/pull/422 JDC logs on this error should become:

```
2026-04-09T00:41:01.708246Z  INFO jd_client_sv2: Job declarator client starting... setting up subsystems, User Identity: your_username_here
2026-04-09T00:41:01.708440Z  INFO jd_client_sv2: Initializing monitoring server on http://0.0.0.0:9091
2026-04-09T00:41:01.708841Z  INFO stratum_apps::monitoring::http_server: Starting monitoring server on http://0.0.0.0:9091
2026-04-09T00:41:01.708854Z  INFO stratum_apps::monitoring::http_server: Cache refresh interval: 15s
2026-04-09T00:41:01.708902Z  INFO jd_client_sv2: Using Bitcoin Core IPC socket at: /Users/plebhash/Library/Application Support/Bitcoin/node.sock
2026-04-09T00:41:01.709211Z  INFO jd_client_sv2: Attempting to initialize upstream...
2026-04-09T00:41:01.709216Z  INFO jd_client_sv2: Trying upstream 1 of 1: pool=75.119.150.111:3333, jds=75.119.150.111:3334
[...]

PR #88, opened 2026-04-09
bootstrap log parsing infra

plebhash:
foundations for subissues of #85

### DIFF
--- a/server/src/docker.ts
+++ b/server/src/docker.ts
@@ -208,6 +208,8 @@ function splitLogLines(
     .split(/\r?\n/)
     .filter((line) => line.length > 0)
     .map((raw) => {
+      // Dockerode prefixes lines with an RFC 3339 timestamp when
+      // `timestamps: true` is enabled, so we split it from the log message.
       const match = raw.match(/^(\d{4}-\d{2}-\d{2}T\S+?)\s(.*)$/);
 
       return {
--- a/server/src/docker.ts
+++ b/server/src/docker.ts
@@ -6,6 +6,7 @@ import fs from 'fs';
 import Docker from 'dockerode';
 import os from 'os';
 import type { SetupData, ContainerStatus, HealthStatus } from './types.js';
+import type { ContainerLogLine, LogContainerRole, LogOutputStream } from './logs/types.js';
 
 /**
  * Expand ~ to home directory in a path.
@@ -146,6 +147,7 @@ const TRANSLATOR_CONTAINER = 'sv2-translator';
 const JDC_CONTAINER = 'sv2-jdc';
 const TRANSLATOR_IMAGE = 'stratumv2/translator_sv2:main';
 const JDC_IMAGE = 'stratumv2/jd_client_sv2:main';
+const DOCKER_LOG_HEADER_SIZE = 8;
 
 /**
  * Detect if we're running inside a Docker container.
@@ -156,6 +158,93 @@ const isRunningInDocker = fs.existsSync('/.dockerenv');
 export function getDockerConnectionInfo(): DockerConnectionConfig {
   refreshDockerConnection();
   return dockerConnection;
+}
+
+const LOG_CONTAINER_NAMES: Record<LogContainerRole, string> = {
+  translator: TRANSLATOR_CONTAINER,
+  jdc: JDC_CONTAINER,
+};
+
+type DockerLogChunk = {
+  stream: LogOutputStream;
+  payload: string;
+};
+
+function demuxDockerLogBuffer(buffer: Buffer): DockerLogChunk[] {
+  const chunks: DockerLogChunk[] = [];
+  let offset = 0;
+
+  while (offset + DOCKER_LOG_HEADER_SIZE <= buffer.length) {
+    const streamType = buffer.readUInt8(offset);
+    const payloadLength = buffer.readUInt32BE(offset + 4);
+    const payloadStart = offset + DOCKER_LOG_HEADER_SIZE;
+    const payloadEnd = payloadStart + payloadLength;
+
+    if (payloadEnd > buffer.length) {
+      break;
+    }
+
+    chunks.push({
+      stream: streamType === 2 ? 'stderr' : 'stdout',
+      payload: buffer.subarray(payloadStart, payloadEnd).toString('utf-8'),
+    });
+
+    offset = payloadEnd;
+  }
+
+  if (chunks.length === 0 && buffer.length > 0) {
+    return [{ stream: 'stdout', payload: buffer.toString('utf-8') }];
+  }
+
+  return chunks;
+}
+
+function splitLogLines(
+  container: LogContainerRole,
+  stream: LogOutputStream,
+  payload: string
+): ContainerLogLine[] {
+  return payload
+    .split(/\r?\n/)
+    .filter((line) => line.length > 0)
+    .map((raw) => {
+      const match = raw.match(/^(\d{4}-\d{2}-\d{2}T\S+?)\s(.*)$/);
+
+      return {
+        container,
+        stream,
+        timestamp: match ? match[1] : null,
+        message: match ? match[2] : raw,
+        raw,
+      };
+    });
+}
+
+export async function readContainerLogs(
+  container: LogContainerRole,
+  options: { tail?: number } = {}
+): Promise<ContainerLogLine[]> {
+  refreshDockerConnection();
+
+  try {
+    const dockerContainer = docker.getContainer(LOG_CONTAINER_NAMES[container]);
+    const info = await dockerContainer.inspect();
+    const logBuffer = await dockerContainer.logs({
+      stdout: true,
+      stderr: true,
+      timestamps: true,
+      tail: options.tail,
+      abortSignal: AbortSignal.timeout(2000),
+    });
+
+    const chunks = info.Config?.Tty
+      ? [{ stream: 'stdout' as const, payload: logBuffer.toString('utf-8') }]
+      : demuxDockerLogBuffer(logBuffer);
+
+    return chunks.flatMap((chunk) => splitLogLines(container, chunk.stream, chunk.payload));
+  } catch {
+    return [];
+  }
 }
 
 /**
--- a/server/src/docker.ts
+++ b/server/src/docker.ts
@@ -244,8 +244,10 @@ export async function readContainerLogs(
       : demuxDockerLogBuffer(logBuffer);
 
     return chunks.flatMap((chunk) => splitLogLines(container, chunk.stream, chunk.payload));
-  } catch {
-    return [];
+  } catch (error) {
+    throw new Error(`Failed to read logs for ${container} container`, {
+      cause: error instanceof Error ? error : new Error(String(error)),
+    });
   }
 }
 
--- a/server/src/docker.ts
+++ b/server/src/docker.ts
@@ -170,6 +170,8 @@ type DockerLogChunk = {
   payload: string;
 };
 
+// Docker uses an 8-byte framing header for non-TTY stdout/stderr multiplexing.
+// Reference: https://docs.docker.com/reference/api/engine/version/v1.45/#tag/Container/operation/ContainerAttach
 function demuxDockerLogBuffer(buffer: Buffer): DockerLogChunk[] {
   const chunks: DockerLogChunk[] = [];
   let offset = 0;
--- a/server/src/index.ts
+++ b/server/src/index.ts
@@ -20,6 +20,7 @@ import {
   ensureDockerAvailable,
   getDockerConnectionInfo,
 } from './docker.js';
+import { getLogDiagnostics } from './logs/diagnostics.js';
 
 const __dirname = path.dirname(fileURLToPath(import.meta.url));
 const app = express();
@@ -121,6 +122,20 @@ app.get('/api/config', async (_req, res) => {
   } catch (error) {
     console.error('Config error:', error);
     res.status(500).json({ error: 'Failed to get config' });
+  }
+});
+
+/**
+ * GET /api/logs/diagnostics - Get collated log diagnostics for the deployed stack
+ */
+app.get('/api/logs/diagnostics', async (_req, res) => {
+  try {
+    const state = await loadState();
+    const response = await getLogDiagnostics(state.mode, state.configured);
+    res.json(response);
+  } catch (error) {
+    console.error('Log diagnostics error:', error);
+    res.status(500).json({ error: 'Failed to get log diagnostics' });
   }
 });
 
--- a/server/src/logs/parsers.ts
+++ b/server/src/logs/parsers.ts
@@ -1,0 +1,12 @@
+import type { ContainerLogLine, LogDiagnostic, LogParser } from './types.js';
+
+export const logParsers: LogParser[] = [];
+
+export function collectDiagnostics(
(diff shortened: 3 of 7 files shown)


plebhash:
as everyone knows, I'm not a TS dev so this was 100% written by `gpt-5.4-codex`

context came from the github issues + current codebase @ 0a6ea9f904210464f1f95ec60719875604d88971

more specifically, we're trying to lay the foundations for:
- https://github.com/stratum-mining/sv2-ui/issues/82#issuecomment-4207061453
- https://github.com/stratum-mining/sv2-ui/issues/82#issuecomment-4207203806
- #83 
- #86 
- #87

(although this is NOT trying to solve #82 #86 or #87 specifically, as I think we should tackle those as a follow-up PRs with more well-delimited scopes)

so a sanity check by human eyes (skilled in TS/JS+HTML+CSS) should be done before we merge this

---

TLDR: the goal here is to establish a sane foundation for log-based errors (from multiple different scenarios) on the UI moving forward

server/src/docker.ts
```diff
@@ -158,6 +160,93 @@ export function getDockerConnectionInfo(): DockerConnectionConfig {
...
+function splitLogLines(
+  container: LogContainerRole,
+  stream: LogOutputStream,
+  payload: string
+): ContainerLogLine[] {
+  return payload
+    .split(/\r?\n/)
+    .filter((line) => line.length > 0)
+    .map((raw) => {
+      const match = raw.match(/^(\d{4}-\d{2}-\d{2}T\S+?)\s(.*)$/);
```
lucasbalieiro, line 215: are we trying to match a timestamp here?

If yes, could you write a comment on top of it?

lucasbalieiro (commented):
well, I did the first superficial review: 3b9a979c67e2b917f4fc813e39fc52ae94f7de5f

The file structure seems sane and is following the existing partterns of the codebase.

I'm still wrapping head around the code because there is too many indirections, it is hard to follow when you don't have a concrete case.

the endpoint is up and running:

Image


The app is able to get the logs from the container:

Image


I still need to check scenarios of: 
-  how the buffer size of the logs grows overtime (because the tail is optional).
- check scenarios with JDC
- check scenarios with `panic!`s

server/src/docker.ts
```diff
@@ -158,6 +160,93 @@ export function getDockerConnectionInfo(): DockerConnectionConfig {
...
+      abortSignal: AbortSignal.timeout(2000),
+    });
+
+    const chunks = info.Config?.Tty
+      ? [{ stream: 'stdout' as const, payload: logBuffer.toString('utf-8') }]
+      : demuxDockerLogBuffer(logBuffer);
+
+    return chunks.flatMap((chunk) => splitLogLines(container, chunk.stream, chunk.payload));
+  } catch {
+    return [];
```
lucasbalieiro: I would recommend trowing an error here instead of an empty array.

Because it will be harder to debug when an error occurs while reading the logs from the container, since it will silently swallow failures.

server/src/docker.ts
plebhash: 4fa684aab925bf863760832f69076b3d89b3de5d

server/src/docker.ts
plebhash: 9c99db64001bf9c010bbbce7103aca5c2ecf5aaa

server/src/index.ts
```diff
@@ -124,6 +125,20 @@ app.get('/api/config', async (_req, res) => {
   }
 });
 
+/**
+ * GET /api/logs/diagnostics - Get collated log diagnostics for the deployed stack
+ */
+app.get('/api/logs/diagnostics', async (_req, res) => {
```
lucasbalieiro, line 131: I’m not saying this will be a problem, but there’s an important limitation to keep in mind:

with the current approach, logs are only fetched from the container when the frontend calls this endpoint. As a result, we only ever return the latest ~200 lines at the time of the request.

This means that if the frontend last queried the endpoint, and more than 200 log lines were generated afterward, any earlier errors will no longer be visible. For example, if an error occurs at line 201 after the last fetch, it won’t be included in the response.

In practice, this likely isn’t an issue for most of our target scenarios, since many of the errors we care about prevent the application from starting. For example:

* The `unknown-user` case (Braiins): if the channel doesn’t open, no further logs are produced, so the error remains within the latest 200 lines.
* `InvalidCertificate` due to an outdated system clock: the TLS handshake fails and the system never fully initializes.
* Missing `node.sock`: similarly prevents startup.

However, we may want to consider cases where the system continues running after an error. For instance, if the upstream becomes unavailable and JDC enters fallback mode, it may recover and continue operating normally. In that case, the fallback event could be pushed out of the last 200 lines before the frontend queries this endpoint, making it invisible to the user.

This is probably acceptable for now, especially since this PR is focused on establishing the plumbing. But it may be worth addressing in a follow-up by continuously collecting logs from the container and persisting relevant diagnostics, so they can be returned regardless of when the frontend queries the endpoint.

server/src/logs/parsers.ts
lucasbalieiro, line 1: I think i understood the logic here:

Later, after this PR, we will have to implement a `LogParser` for each scenario we want to give a diagnostic.

Example, we might create one the look in the logs and search for the `unknown-user` string and then show a diagnostic with something like: `the $username is not a valid Braiins user, fix by ...`.

I think would be nice to add a deduplication logic here, because we have retry polices for some cases. So we would see duplicated diagnostics since we could have multiple logs pointing to the same error.

server/src/docker.ts
```diff
@@ -158,6 +160,97 @@ export function getDockerConnectionInfo(): DockerConnectionConfig {
...
+      stdout: true,
+      stderr: true,
+      timestamps: true,
+      tail: options.tail,
+      abortSignal: AbortSignal.timeout(2000),
+    });
+
+    const chunks = info.Config?.Tty
+      ? [{ stream: 'stdout' as const, payload: logBuffer.toString('utf-8') }]
+      : demuxDockerLogBuffer(logBuffer);
```
lucasbalieiro, line 246: I have a blindspot here, need a bit more of time to investigate why this `Tty` assertion is needed.

server/src/index.ts
plebhash: I see two paths forward:

1. document `RECENT_LOG_TAIL` making it clear that it's an arbitrary value that could have limiting consequences on specific edge cases
2. expand `RECENT_LOG_TAIL` to higher values, while making sure they don't make it prohibitive to run on cheap consumer-grade hardware

we can probably start with 1. (17ed49bedcf01e09e4a1c66e1389969c43a8cfbc) and then dive into 2. if/when necessary

wdyt @lucasbalieiro ?

server/src/logs/parsers.ts
plebhash: 03148bdb116d0032de892859c4101759bc191b9c

server/src/docker.ts
lucasbalieiro: Yeah, the assertion makes sense. 

TIL. I didn't know that we could create a container with tty enabled. It won't make a difference for us because the code that is starting the container will use the default: `tty=false`. Which will always fall in the multiplexed logs. But I would keep this assertion anyway.

server/src/index.ts
lucasbalieiro: Let's go 1. Ack

server/src/docker.ts
plebhash: according to my clanker:

> if TTY is disabled, Docker multiplexes stdout/stderr with its 8-byte framing header, so we need to demux the buffer.
>
> if TTY is enabled, Docker returns one combined raw stream instead, so demuxing would be incorrect and we should treat the whole buffer as a single stream.

lucasbalieiro: > according to my clanker:
> 
> > if TTY is disabled, Docker multiplexes stdout/stderr with its 8-byte framing header, so we need to demux the buffer.
> > if TTY is enabled, Docker returns one combined raw stream instead, so demuxing would be incorrect and we should treat the whole buffer as a single stream.

and he is right, if you want, maybe add a comment linking the docs. So our future selfs don't try to optimize it without context:

https://docs.docker.com/reference/api/engine/version/v1.45/#tag/Container/operation/ContainerAttach

this link has the HEADER format, giving some rationale behind the parsing logic

server/src/logs/parsers.ts
lucasbalieiro: cACK

server/src/docker.ts
plebhash: 1e0ea28

src/types/log-diagnostics.ts
```diff
@@ -0,0 +1,39 @@
...
+  id: 'mining-services';
+  label: string;
+  containers: LogContainerRole[];
+  collated: boolean;
+  source: LogSourceKind;
+}
+
+export interface LogDiagnosticsResponse {
+  configured: boolean;
+  mode: SetupMode | null;
```
bit-aloo, line 35: What a null mode suggests? shouldn't setupMode have a default value in case not specified?

src/types/log-diagnostics.ts
```diff
@@ -0,0 +1,39 @@
...
+
+export interface LogDiagnostic {
+  code: string;
+  severity: 'warning' | 'error';
+  title: string;
+  message: string;
+  recommendation: string;
+  streamId: 'mining-services';
+  containers: LogContainerRole[];
+  detectedAt: string | null;
```
bit-aloo, line 21: Shouldn't this be `Date`?

src/types/log-diagnostics.ts
```diff
@@ -0,0 +1,39 @@
+export type LogContainerRole = 'translator' | 'jdc';
+export type LogOutputStream = 'stdout' | 'stderr';
+export type LogSourceKind = 'docker-container-logs' | 'container-log-file';
+export type SetupMode = 'jd' | 'no-jd';
+
+export interface DiagnosticEvidence {
+  container: LogContainerRole;
+  stream: LogOutputStream;
+  timestamp: string | null;
```
bit-aloo, line 9: We can make it `Date`

server/src/index.ts
```diff
@@ -124,6 +125,20 @@ app.get('/api/config', async (_req, res) => {
...
+  try {
+    const state = await loadState();
+    const response = await getLogDiagnostics(state.mode, state.configured);
+    res.json(response);
+  } catch (error) {
+    console.error('Log diagnostics error:', error);
+    res.status(500).json({ error: 'Failed to get log diagnostics' });
+  }
+});
```
bit-aloo, line 140: I was under the impression that we wanna stream data from container, this looks more like polling?

bit-aloo, line 140: I was under the impression that we wanna stream data from container, this looks more like polling?

src/types/log-diagnostics.ts
```diff
@@ -0,0 +1,39 @@
+export type LogContainerRole = 'translator' | 'jdc';
+export type LogOutputStream = 'stdout' | 'stderr';
+export type LogSourceKind = 'docker-container-logs' | 'container-log-file';
+export type SetupMode = 'jd' | 'no-jd';
+
```
bit-aloo, line 5: Shouldn't this file be inside types.ts?

src/types/log-diagnostics.ts
plebhash: according to my clanker:

> null means “not configured yet” / “no persisted setup mode exists yet”. I don’t think this should have a default value, because that would invent configuration that the user has not actually chosen.`

server/src/index.ts
plebhash: according to my clanker:

> Yes, this endpoint is polling/snapshot-based by design. It reads the current tail from the container at request time and returns structured diagnostics for that snapshot. Streaming logs continuously would be a different implementation choice; this PR is only establishing the diagnostics API/parsing foundation, not a streaming transport.

lucasbalieiro (commented):
tACK 1e0ea28e8498768d4df9a51a5e64483b12256bc8 with multiple containers.

And when container state is restarting/exited. Also worked successfully

Image

plebhash:
@lucasbalieiro can you approve the PR then?

(as in github PR review approve)

lucasbalieiro (approved):
well, I have more firm grasp of what this code is doing now. 

I think further refactoring that I have in mind should only be made when the concrete use case is implemented.

As we discussed the polling limitations (https://github.com/stratum-mining/sv2-ui/pull/88#discussion_r3060248228), I saw that @Shourya742 made a similar question. If you guys decide that we should change to streaming I drop my ACK, so I can re-review it

plebhash:
personally, I'm not too worried about limitations around `RECENT_LOG_TAIL` for the short term:
- for now I cannot think of cases where we need more than 200 of log lines to determine what kind of error has happened
- if needed, expanding `RECENT_LOG_TAIL` to slightly larger values (500 or 1000) should be acceptable, without significant performance impact

so I think we're good for #85

but I do agree that streaming would be more appropriate design for the scope of #82, because that would give a more natural/live feel for how the logs are displayed in the UI (log lines appear as they are generated on the process, not in delayed bursts of multiple lines)

however, #82 feels not so high priority, given that it's a feature for advanced users and nobody has really asked for it yet

so I guess we have two options:
- continue with the poll-based approach for now (which @lucasbalieiro has already reviewed and ACKd)
- pivot this PR to a stream-based design

I asked my clanker to evaluate the effort for pivoting towards a stream-based design:
> this would not be a full rewrite, but it is more than just a small tweak.
>
> the parser/dedup layer should survive mostly intact, but we would need to replace the one-shot log fetch with long-lived log-follow collection, add a streaming transport to the backend (likely SSE or WebSocket), and update the frontend hook/consumer model accordingly.
>
> so it feels like a meaningful transport/collection redesign, not something I would casually fold into this PR unless we want to lay more solid foundations for #82 sooner rather than later.

@lucasbalieiro which direction you feel is the best to take?

lucasbalieiro:
> personally, I'm not too worried about limitations around `RECENT_LOG_TAIL` for the short term:
> - for now I cannot think of cases where we need more than 200 of log lines to determine what kind of error has happened
> - if needed, expanding `RECENT_LOG_TAIL` to slightly larger values (500 or 1000) should be acceptable, without significant performance impact
> 
> so I think we're good for #85
> 
> but I do agree that streaming would be more appropriate design for the scope of #82, because that would give a more natural/live feel for how the logs are displayed in the UI (log lines appear as they are generated on the process, not in delayed bursts of multiple lines)
> 
> however, #82 feels not so high priority, given that it's a feature for advanced users and nobody has really asked for it yet
> 
> so I guess we have two options:
> - continue with the poll-based approach for now (which @lucasbalieiro has already reviewed and ACKd)
> - pivot this PR to a stream-based design
> 
> I asked my clanker to evaluate the effort for pivoting towards a stream-based design:
> > this would not be a full rewrite, but it is more than just a small tweak.
> >
> > the parser/dedup layer should survive mostly intact, but we would need to replace the one-shot log fetch with long-lived log-follow collection, add a streaming transport to the backend (likely SSE or WebSocket), and update the frontend hook/consumer model accordingly.
> >
> > so it feels like a meaningful transport/collection redesign, not something I would casually fold into this PR unless we want to lay more solid foundations for #82 sooner rather than later.
> 
> @lucasbalieiro which direction you feel is the best to take?

Let's go with 1. I'll implement the stream later once we need it

lucasbalieiro (approved):

PR #90, opened 2026-04-10
add socket path validation during setup

rx18-eng:
## Title
add socket path validation and fix jdc bind mount for testnet4

### what this does

when you set up JD mode and the bitcoin core socket path is wrong or doesn't exist, the wizard used to just let you through and then JDC would crash-loop with a confusing "cannot connect to pool" error on the dashboard. you had no idea the actual problem was the socket path.

this adds a validation check on the bitcoin core setup step — it hits the backend to verify the socket exists before you can continue. if its missing you get a clear message like "socket not found at /path — make sure bitcoin core is running with IPC enabled" and the continue button is blocked.

### changes

- server/src/docker.ts — exported expandHomePath
- server/src/index.ts — added POST /api/validate/bitcoin-socket endpoint
- src/components/setup/steps/BitcoinSetup.tsx — added debounced socket validation with inline status display

### test plan

- go through setup with JD mode on testnet4 with bitcoin core running — should show green "socket found" and complete setup
- try with a wrong path — should show red error and block continue
- check that JDC container actually connects to the socket after setup (no more crash-loop)

partially address #83

https://github.com/user-attachments/assets/9466869e-c3cc-425a-b706-509083b6ae98

### DIFF
--- a/server/src/docker.ts
+++ b/server/src/docker.ts
@@ -13,7 +13,7 @@ import type { ContainerLogLine, LogContainerRole, LogOutputStream } from './logs
  * Uses HOST_HOME env var (passed from docker run) if available,
  * otherwise falls back to os.homedir() (works in development).
  */
-function expandHomePath(inputPath: string): string {
+export function expandHomePath(inputPath: string): string {
   if (inputPath.startsWith('~')) {
     const home = process.env.HOST_HOME || os.homedir();
     return inputPath.replace('~', home);
--- a/server/src/index.ts
+++ b/server/src/index.ts
@@ -6,6 +6,7 @@
 
 import express from 'express';
 import cors from 'cors';
+import net from 'net';
 import path from 'path';
 import fs from 'fs/promises';
 import { fileURLToPath } from 'url';
@@ -19,6 +20,7 @@ import {
   isDockerAvailable,
   ensureDockerAvailable,
   getDockerConnectionInfo,
+  expandHomePath,
 } from './docker.js';
 import { getLogDiagnostics } from './logs/diagnostics.js';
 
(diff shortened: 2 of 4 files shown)


rx18-eng:
@pavlenex @plebhash  Please give it a look on this !

server/src/docker.ts
```diff
@@ -320,18 +320,25 @@ async function startTranslator(configPath: string): Promise<void> {
...
   configPath: string,
-  bitcoinSocketPath: string
+  bitcoinSocketPath: string,
+  network: string
 ): Promise<void> {
   await removeContainer(JDC_CONTAINER);
 
+  // JDC resolves the socket path based on network: /root/.bitcoin/node.sock for mainnet,
+  // /root/.bitcoin/testnet4/node.sock for testnet4. Mount to the path JDC actually expects.
+  const containerSocketPath = network === 'mainnet'
```
lucasbalieiro, line 435: please, create a separate commit for this fix of the `testnet4` error. Also, you can reference the issue on your PR description. 

I've created this issue for this bug:
https://github.com/stratum-mining/sv2-ui/issues/106

also, thanks for reporting

server/src/index.ts
```diff
@@ -124,6 +125,47 @@ app.get('/api/config', async (_req, res) => {
...
+app.post('/api/validate/bitcoin-socket', async (req, res) => {
+  const { socket_path } = req.body;
+  if (!socket_path || typeof socket_path !== 'string') {
+    return res.status(400).json({ valid: false, error: 'socket_path is required' });
+  }
+
+  const resolved = expandHomePath(socket_path);
+
+  try {
+    const stat = await fs.stat(resolved);
```
lucasbalieiro: I got your instinct here, but checking the filesystem could lead to false positives.

When you start a bitcoin node with IPC enable it creates the `node.sock` file. But once you stop the node, the file still there. 

As an example in the image below, you can see that I stopped my container and the validation still gives green light in the setup process:

Image


I would recommend another approach, try to really connect to the socket, node has the capability to deal with IPC connections:

https://nodejs.org/docs/latest-v20.x/api/net.html#socketconnectpath-connectlistener


If the connection succeed we can have more assurance that the bitcoin core not only created the file but is actually on and listening the socket.

But if the connection fails it will probably throw an `ENOENT` if the file does not exists or `ECONNREFUSED` if the file exists but the node.sock is not attached to any process. The specific errors needs to be checked for each scenario, but I think you get the idea

https://nodejs.org/docs/latest-v20.x/api/errors.html#common-system-errors

src/components/setup/steps/BitcoinSetup.tsx
```diff
@@ -20,10 +22,47 @@ export function BitcoinSetup({ data, updateData, onNext }: StepProps) {
...
+        setSocketError(result.error || 'Socket validation failed');
+      }
+    } catch {
+      // Server not reachable — skip validation (standalone mode)
+      setSocketStatus('idle');
+    }
+  }, []);
+
+  // Debounced validation when socket path changes
+  useEffect(() => {
```
lucasbalieiro: why not use `react-query` here?

src/components/setup/steps/BitcoinSetup.tsx
```diff
@@ -151,13 +154,33 @@ export function BitcoinSetup({ data, updateData, onNext }: StepProps) {
...
+            <CheckCircle2 className="h-4 w-4" />
+            <span className="text-xs">Socket is listening</span>
+          </div>
+        )}
+        {!isChecking && socketError && (
+          <div className="flex items-start gap-2 mt-3 text-sv2-red">
+            <AlertCircle className="h-4 w-4 flex-shrink-0 mt-0.5" />
+            <span className="text-xs">{socketError}</span>
+          </div>
+        )}
```
lucasbalieiro, line 190: please, add some accent colors here. Maybe green and red for errors and successes.

lucasbalieiro: Image

lucasbalieiro:
@rx18-eng , thanks. now this is in the right direction

@pavlenex, you might want to take a look at the design choices/erros messages

rx18-eng:
Screenshot 2026-04-17 004646
Screenshot 2026-04-17 004633

pavlenex:
Can we make design consistent with what we have in terms of colors and ? We cannot allow AI to push random design with every PR, because things are going to get out of hand.
 
Screenshot 2026-04-17 at 16 15 49

pavlenex:
Same goes for success message!

pavlenex:
@lucasbalieiro Do we have validation for mainnet? I am unsure why code is fragmenting there.

lucasbalieiro:
> @lucasbalieiro Do we have validation for mainnet? I am unsure why code is fragmenting there.

yes, the socket validation works for mainnet and testnet4

rx18-eng:
@lucasbalieiro @pavlenex please give it a look on this !

pavlenex (approved):
tACK

lucasbalieiro (approved):
tACK ef1775dbb0b0fa8fd89804f47cf08cd0f7087448

PR #96, opened 2026-04-11
show diagnostic for `unknown-user` when connecting to Braiins pool

### LINKED ISSUE #86: bad braiins username should show specific error

lucasbalieiro:
closes #86 
lay the foundations for #21 

a few changes were needed beyond implementing the parser:

1. updated the `RestartPolicy` of the tproxy container. With automatic restarts, the diagnostic banner would repeatedly blink as the container restarted. The container now stops on failure, since it cannot fully initialize in this state.

2. Adjusted log parsing to consider only entries from the current container start time onward. This avoids false diagnostics caused by logs from previous runs. For example, if a prior run failed due to an unknown user but was later fixed, the old error logs could still trigger the diagnostic within the 200-line threshold.

image

--

togive the user the ability fix the `unknown-user` error, which requires correcting `user_identity`, I also started groundwork for #21:

1. added a `PUT` endpoint to accept partial config updates and apply them where needed.
2. built a reusable inline editing component.
3. updated the dashboard to support inline editing of the pool username, avoiding full reconfiguration.

image

### DIFF
--- a/server/src/docker.ts
+++ b/server/src/docker.ts
@@ -277,23 +277,23 @@ async function connectSv2UiToNetwork(): Promise<void> {
   try {
     // Find sv2-ui container (could be named sv2-ui or sv2-ui-test)
     const containers = await docker.listContainers({ all: true });
-    const sv2UiContainer = containers.find(c => 
+    const sv2UiContainer = containers.find(c =>
       c.Names.some(n => n === '/sv2-ui' || n === '/sv2-ui-test')
     );
-    
+
     if (!sv2UiContainer) {
       // Not running in Docker (development mode)
       return;
     }
-    
+
     const network = docker.getNetwork(NETWORK_NAME);
     const networkInfo = await network.inspect();
-    
+
     // Check if already connected
     if (networkInfo.Containers && networkInfo.Containers[sv2UiContainer.Id]) {
       return;
     }
-    
+
     console.log('Connecting sv2-ui to sv2-network...');
     await network.connect({ Container: sv2UiContainer.Id });
   } catch {
@@ -345,7 +345,7 @@ async function getContainerStatus(name: string): Promise<ContainerStatus | null>
   try {
     const container = docker.getContainer(name);
     const info = await container.inspect();
-    
+
     let status: HealthStatus = 'stopped';
     if (info.State.Running) {
       status = info.State.Health?.Status === 'healthy' ? 'healthy' : 'starting';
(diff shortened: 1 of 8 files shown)


plebhash:
I was able to reproduce the `Invalid Braiins username` error card

but it took a while to load... for a few seconds, it was showing the red card with `Cannot connect to pool. Make sure mining services are running.` 

so I was asking myself: "is everything ok here?"

I wonder if that's because of this line?
https://github.com/stratum-mining/sv2-ui/blob/1714d5164031ea161e18cbf67e8524c4c397e841/src/hooks/useLogDiagnostics.ts#L24

if yes, I wonder if we can reduce it to a smaller value to make this more responsive?

plebhash:
please note that the context for my comment above was related to the fact that I wasn't starting from a fresh setup

instead, I was just hitting `Reconfigure`.

my clanker suggested this, which helped make it more responsive:
```diff
diff --git a/src/hooks/useControlApi.ts b/src/hooks/useControlApi.ts
index 8c12de7..1c2c868 100644
--- a/src/hooks/useControlApi.ts
+++ b/src/hooks/useControlApi.ts
@@ -61,6 +61,7 @@ export function useControlApi() {
     onSuccess: () => {
       // Invalidate status queries
       queryClient.invalidateQueries({ queryKey: ['setup-status'] });
+      queryClient.invalidateQueries({ queryKey: ['log-diagnostics'] });
     },
   });
 
@@ -68,6 +69,7 @@ export function useControlApi() {
     mutationFn: restartServices,
     onSuccess: () => {
       queryClient.invalidateQueries({ queryKey: ['setup-status'] });
+      queryClient.invalidateQueries({ queryKey: ['log-diagnostics'] });
     },
   });
 
@@ -75,6 +77,7 @@ export function useControlApi() {
     mutationFn: setupServices,
     onSuccess: () => {
       queryClient.invalidateQueries({ queryKey: ['setup-status'] });
+      queryClient.invalidateQueries({ queryKey: ['log-diagnostics'] });
     },
   });
 
@@ -82,6 +85,7 @@ export function useControlApi() {
     mutationFn: updateConfigService,
     onSuccess: () => {
       queryClient.invalidateQueries({ queryKey: ['setup-status'] });
+      queryClient.invalidateQueries({ queryKey: ['log-diagnostics'] });
     },
   });
```

lucasbalieiro:
@plebhash , thanks. I forgot about the `useMutation` thing.

[078cfdd](https://github.com/stratum-mining/sv2-ui/pull/96/commits/078cfdd849453a97213ad77e399a5b2b65cac1d6)

Reduced the poll time.

plebhash (approved):
tACK

PR #98, opened 2026-04-12
fix: polish light mode component visibility

### LINKED ISSUE #97: light mode components needs polishing
Some buttons and components disappear when on light mode

Image

When light mode on:

Image

---
Image

when light mode on:

Image
---
Image

when light mode on:

Image

---
Image

When light mode on:

Image

rx18-eng:
closes #97 

## Body

so right now a bunch of components are basically invisible in light mode because the UI was built dark-mode-first. cards, borders, config rows, the chart - all disappear on white.

the root cause is the light mode CSS variables (--card, --border, --muted) are too close to pure white, so anything using partial opacity like bg-card/40 just produces zero contrast. this shifts them slightly darker and removes border-none from glass cards. also bumps opacity values across the board — worker tables, config rows, search bar, nav, popovers.

the other fix is for the configurable accent color. there was no contrast safety, so picking a light yellow as primary made buttons and text vanish. added lightness clamping in useUiConfig that keeps things readable in both modes.

also added missing sv2-green/yellow/red color definitions that badge variants were referencing but were never defined.

### test plan

- toggle light mode, check dashboard cards/chart/table have visible borders
- settings > appearance, pick a very light color, check buttons stay readable
- check FAQ and setup wizard borders visible in light mode
- toggle dark mode, confirm no regressions

https://github.com/user-attachments/assets/46bc9a10-547b-458c-bede-b9cacc49fa2b

### DIFF
--- a/src/hooks/useUiConfig.ts
+++ b/src/hooks/useUiConfig.ts
@@ -39,33 +39,56 @@ function saveConfig(config: UiConfig) {
   window.localStorage.setItem(STORAGE_KEY, JSON.stringify(config));
 }
 
+/**
+ * Clamp a primary color's lightness so it has enough contrast against
+ * the page background in both light and dark mode.
+ */
+function contrastSafeHsl(h: string, s: string, lVal: number, isDark: boolean): { hsl: string; fgIsWhite: boolean } {
+  let l = lVal;
+  if (isDark) {
+    // Too dark on a black background — boost lightness
+    if (l < 30) l = 30;
+  } else {
+    // Too light on a white background — darken
+    if (l > 60) l = 60;
+  }
+  // Foreground: white for dark primaries, black for light ones
+  const fgIsWhite = l < 55;
+  return { hsl: `${h} ${s} ${l}%`, fgIsWhite };
+}
+
 // Apply runtime CSS variable overrides based on config.
-// Slightly boosts lightness for dark mode primary (+5% L).
+// Adjusts lightness for both light and dark mode to keep contrast safe.
 function applyCssVariables(config: UiConfig) {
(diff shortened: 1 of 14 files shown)


rx18-eng:
@lucasbalieiro @pavlenex Please give it alook on this !

src/hooks/useUiConfig.ts
```diff
@@ -39,50 +39,86 @@ function saveConfig(config: UiConfig) {
   window.localStorage.setItem(STORAGE_KEY, JSON.stringify(config));
 }
 
+/**
+ * Clamp a primary color's lightness so it has enough contrast against
+ * the page background in both light and dark mode.
+ */
+function contrastSafeHsl(h: string, s: string, lVal: number, isDark: boolean): { hsl: string; fgIsWhite: boolean } {
```
lucasbalieiro, line 46: this is clever.

Do you have some rationale behind the thresholds or you did it more on a feeling?

lucasbalieiro (commented):
could you rebase?

rx18-eng:
rebased! and for the thresholds - bit of both honestly. 60 cap in light mode cause anything lighter than that on white starts losing contrast fast. 30 floor in dark mode same idea flipped. 55 for the fg switch is just where white text starts working better than black on the primary. tested with a bunch of different accent colors and these held up well but nothing too rigid about the exact numbers

lucasbalieiro (approved):

rx18-eng:
@pavlenex  shall i update anything else on this pr?

pavlenex:
I'm not going to merge this PR just yet, as I already mentioned in the issue itself, the way this solves the problem is odd.

PR #100, opened 2026-04-14
fix docker log regression

plebhash:
fix this regression after merging #96:

https://github.com/stratum-mining/sv2-ui/actions/runs/24415098172/job/71322020516

also adds extra CI steps to prevent PRs from introducing similar regressions in the future


lucasbalieiro (approved):
thanks, tACK 9da058d62e27108a1b0b3d5384adc5d79ec7d70c

PR #107, opened 2026-04-15
move support page next to theme switch, redesign

pavlenex:
This PR moves support page on the right header side next to the theme switch, making it less prominent. It also removes exessive icons on support page and removes bottom text since Join discord call to action seemed too excessive with just two FAQ, we can add that text as we build the page.

I also used SV2 UI instead of SV2 Mining stack, to ensure consistency in how we call it.
Before
Screenshot 2026-04-15 at 23 10 57

After
Screenshot 2026-04-15 at 23 10 44


lucasbalieiro (approved):
tACK d1f5ae87c1b82e97f10b4dfd41974dd7e66a1e85

In a macos + brave browser

Image

PR #112, opened 2026-04-17
add parser for `node.sock` problems

### LINKED ISSUE #83: unavailable `node.sock` error case is displaying misleading error

lucasbalieiro:
closes #83 
to be merged after #90 
I changed the restart policy of the jdc container, so it don't make the message blink in the screen.

added a parser to catch scenarios where the jdc has problems with the `node.sock` connection.

image

### DIFF
--- a/server/src/logs/parsers.test.ts
+++ b/server/src/logs/parsers.test.ts
@@ -1,0 +1,151 @@
+import assert from 'node:assert/strict';
+import test from 'node:test';
+
+import {
+  unknownUserParser,
+  jdcBitcoinCoreDisconnectedParser,
+} from './parsers.js';
+import type { ContainerLogLine } from './types.js';
+
+function createLogLine(
+  container: 'translator' | 'jdc',
+  message: string
+): ContainerLogLine {
+  return {
+    container,
+    stream: 'stderr',
+    timestamp: '2026-04-17T18:30:43.174887Z',
+    message,
+    raw: `2026-04-17T18:30:43.174887Z ERROR ${message}`,
+  };
+}
+
+test('unknownUserParser matches OpenMiningChannelError with unknown-user', () => {
+  const lines = [
+    createLogLine(
+      'translator',
+      'OpenMiningChannelError(request_id: 123, error_code: unknown-user)'
+    ),
+  ];
+
+  const result = unknownUserParser(lines);
+
+  assert.ok(result !== null);
+  assert.equal(result?.code, 'unknown-user');
+  assert.equal(result?.severity, 'warning');
+  assert.equal(result?.title, 'Invalid Braiins username');
+  assert.deepEqual(result?.containers, ['translator']);
+  assert.equal(result?.evidence.length, 1);
+});
+
+test('unknownUserParser returns null when no match', () => {
+  const lines = [
+    createLogLine('translator', 'Some other error message'),
+  ];
+
+  const result = unknownUserParser(lines);
+
+  assert.equal(result, null);
+});
+
+test('unknownUserParser returns null when wrong container', () => {
+  const lines = [
+    createLogLine(
+      'jdc',
+      'OpenMiningChannelError(request_id: 123, error_code: unknown-user)'
+    ),
+  ];
+
+  const result = unknownUserParser(lines);
+
+  assert.equal(result, null);
+});
+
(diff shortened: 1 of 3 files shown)


server/src/logs/parsers.test.ts
```diff
@@ -0,0 +1,151 @@
...
+      'jd_client_sv2::template_receiver::bitcoin_core: Failed to create BitcoinCoreToSv2: CapnpError(Error { kind: Disconnected, extra: "Peer disconnected." })'
+    ),
+  ];
+
+  const result = jdcBitcoinCoreDisconnectedParser(lines);
+
+  assert.ok(result !== null);
+  assert.equal(result?.code, 'jdc-bitcoin-core-disconnected');
+  assert.equal(result?.severity, 'error');
+  assert.equal(result?.title, 'JDC disconnected from Bitcoin Core');
```
plebhash: the user shouldn't need to know what JDC means, we should try to use more generic terminology

server/src/logs/parsers.ts
```diff
@@ -42,10 +45,46 @@ function unknownUserParser(lines: ContainerLogLine[]): LogDiagnostic | null {
...
+      timestamp,
+      line: raw,
+    })
+  );
+
+  return {
+    code: 'jdc-bitcoin-core-disconnected',
+    severity: 'error' as DiagnosticSeverity,
+    title: 'JDC disconnected from Bitcoin Core',
+    message: 'The JDC container lost connection to Bitcoin Core.',
```
plebhash: the user shouldn't need to know what JDC means, we should try to use more generic terminology

plebhash:
I don't have any Bitcoin Core node running locally, and I was able to reproduce the screenshot from the PR description

but I did not see anything on the page where the user provides the Bitcoin Core config (macOS vs linux, custom `node.sock`, etc)

I remember having discussions with @lucasbalieiro around challenges for this.

did we find some blocker? or will that be achieved via some other PR?

lucasbalieiro:
> I don't have any Bitcoin Core node running locally, and I was able to reproduce the screenshot from the PR description
> 
> but I did not see anything on the page where the user provides the Bitcoin Core config (macOS vs linux, custom `node.sock`, etc)
> 
> I remember having discussions with @lucasbalieiro around challenges for this.
> 
> did we find some blocker? or will that be achieved via some other PR?

Another guy already started doing it here: https://github.com/stratum-mining/sv2-ui/pull/90

Since they solve the problem at different stages and with different tools, makes more sense to deal with it separately

server/src/logs/parsers.test.ts
lucasbalieiro: "Bitcoin core stopped working"?

pavlenex:
cACK. I'll leave @plebhash and you to align on the details remaining. i think we can merge this after #90, up to you @lucasbalieiro.

server/src/logs/parsers.test.ts
plebhash: sounds good to me

lucasbalieiro: updated

plebhash:
tACK

image

plebhash (approved):

server/src/logs/parsers.ts
```diff
@@ -42,10 +45,46 @@ function unknownUserParser(lines: ContainerLogLine[]): LogDiagnostic | null {
...
+    })
+  );
+
+  return {
+    code: 'jdc-bitcoin-core-disconnected',
+    severity: 'error' as DiagnosticSeverity,
+    title: 'Bitcoin Core stopped running',
+    message: 'We lost connection with your Bitcoin Core node.',
+    recommendation:
+      'Ensure Bitcoin Core is running and the Unix socket path is correct.',
```
GitGab19: Do we expect users to know what the socket path is?

pavlenex: No, we need to simplify this :)

lucasbalieiro: updated

lucasbalieiro: "Make sure your Bitcoin Core node is up and running on the same computer where you’re running the sv2-ui app."

Issue #70, opened 2026-04-06
consider leveraging `asic-rs`


plebhash:
https://github.com/256foundation/asic-rs

plebhash:
possibly correlated #2

b-rowan:
We would be open to collaborating on this, just would need some details as to how you would want it implemented.  I don't do as much on the frontend, but can add the functionality to the API as needed.

Is there a list of data the team would want to see?

The other consideration is that we don't have any sort of JS bindings, just rust (core) and Python bindings right now, so that would be something to work on.

b-rowan:
@lucasbalieiro

lucasbalieiro:
I'm lacking context about this `asic-rs`. From the their readme seems that this is used to control the asic itself and the `sv2-ui` is to control the `sv2-apps`, totally different jobs.

> Is there a list of data the team would want to see?

@plebhash, did you have anything in mind when you opened this issue? Maybe we could leverage it to get a more accurate hashrate, instead of relying on the `nominal_hashrate` of the channel. but this seems that will change the flow of the app since now we will have to acquire the asics IPs to collect this data. 

Anyhow, I'm unsure on how or for what we should/could use this

> The other consideration is that we don't have any sort of JS bindings, just rust (core) and Python bindings right now, so that would be something to work on.

as for this, we could try some webassembly instead of creating an entire binding just for JS.

plebhash:
ultimately I think @lucasbalieiro will be able to answer the questions above from a more sane pov

but if I had to imagine some integration, I wouldn't necessarily jump into bindings

one key factor is the fact that this project is already doing:
- deployment/orchestration of subapps (currently limited to docker containers)
- frontend communicates with subapps via HTTP

so we could theoretically deploy some executable rust crate that:
- leverages `asic-rs` as a dependency
- communicates with it via HTTP

b-rowan:
> I'm lacking context about this `asic-rs`. From the their readme seems that this is used to control the asic itself and the `sv2-ui` is to control the `sv2-apps`, totally different jobs.

Correct.  `asic-rs` is for miner communication and control.

> Anyhow, I'm unsure on how or for what we should/could use this

My thought on initial look at this was maybe correlating miner temps, power usage, etc in the short term, and possibly adding reboots or power control in the long term?  Not sure what direction would be best.

> as for this, we could try some webassembly instead of creating an entire binding just for JS.

[...]

PR #114, opened 2026-04-18
display container logs for advanced users

### LINKED ISSUE #82: (optionally) display container logs for advanced users
this was discussed/proposed by someone in the recent past (I can't remember who/when)

the idea is to display container logs on the UI for advanced users

it should be an optional thing (not displayed by default)

jayvaliya:
Closes #82

### Summary

Adds a toggleable container logs panel to the bottom of the dashboard. The panel is **hidden by default** and can be enabled via a switch toggle, making it an opt-in feature for advanced users. When enabled in JD mode, logs from both the translator and JDC containers are **collated** into a single view sorted by timestamp, abstracting away the two-container deployment as discussed in #82.

### What changed

**Backend**
- New `GET /api/logs/raw` endpoint that returns collated raw log lines
  - Accepts `?tail=N` query param (default 200, capped at 500)
  - Reuses the existing `readCollatedLogLines` from the log parsing infra (#88)
  - Overrides the internal `RECENT_LOG_TAIL` with the API-supplied tail value
- Exported `readCollatedLogLines` and `getLogStreams` from `diagnostics.ts` for reuse

**Frontend**
- `ContainerLogsPanel` component — terminal-style log viewer with:
  - Auto-scroll to bottom on new lines (pauses when user scrolls up)
  - stderr lines in red, stdout lines in green
  - Container role badges (translator/jdc) shown only in JD mode
  - Download button to export logs as `.txt`
- `useContainerLogs` hook — polls `/api/logs/raw` every 3s when enabled, fully idle when disabled
- `showContainerLogs` boolean added to `useUiConfig` — persisted in localStorage so the toggle survives page reloads
- `Switch` toggle on the dashboard labeled "Container Logs"

**Types**
- Added `ContainerLogLine` and `ContainerLogsResponse` interfaces to `src/types/log-diagnostics.ts` (mirrors the server-side types)

### Screenshots
**Translator only mode (no jd)**
image

**jd mode (translator + jdc collated)**
image

### How it was tested

- **TypeScript compilation**: `tsc --noEmit` passes cleanly for both frontend and server projects
- **Existing tests**: `npx tsx --test server/src/logs/diagnostics.test.ts` — 2/2 pass (missing container handling, error propagation)
- **Lint**: `npx eslint` passes on all changed files
- **Manual testing**:
  - Verified toggle is OFF by default on fresh load (no localStorage entry)
  - Toggled ON → logs appear within 3s, auto-scroll works
  - Scrolled up in the log panel → auto-scroll pauses; scrolled back to bottom → auto-scroll resumes
  - Tested in `no-jd` mode → only translator logs shown, no container badges
  - Tested in `jd` mode → both translator and jdc logs collated, sorted by timestamp, badges visible
  - Clicked "Download logs" → `.txt` file downloaded with correct format (`timestamp [container] [stream] message`)
  - Toggled OFF → polling stops (confirmed via network tab, no more `/api/logs/raw` requests)
  - Refreshed page → toggle state persisted from localStorage
  - Tested with containers stopped → "No log output yet. Services may not be running." message shown

### Design decisions

- **Polling, not streaming**: Consistent with the approach agreed upon in #88. Streaming (SSE/WebSocket) was discussed but deferred as a future enhancement
- **3s poll interval**: Matches the existing diagnostics polling cadence; fast enough for a "live" feel without excessive load
- **localStorage for toggle**: Lightweight persistence without needing server-side state for a UI preference
- **Collated view**: Per plebhash's comment in #82, JDC + tProxy logs are funnelled into one box to abstract away the multi-container deployment

### DIFF
--- a/server/src/index.ts
+++ b/server/src/index.ts
@@ -21,8 +21,9 @@ import {
   ensureDockerAvailable,
   getDockerConnectionInfo,
   expandHomePath,
+  readContainerLogs
 } from './docker.js';
-import { getLogDiagnostics } from './logs/diagnostics.js';
+import { getLogDiagnostics, getLogStreams, readCollatedLogLines } from './logs/diagnostics.js';
 
 const __dirname = path.dirname(fileURLToPath(import.meta.url));
 const app = express();
@@ -284,6 +285,44 @@ app.get('/api/logs/diagnostics', async (_req, res) => {
   } catch (error) {
     console.error('Log diagnostics error:', error);
     res.status(500).json({ error: 'Failed to get log diagnostics' });
+  }
+});
+
+/**
+ * GET /api/logs/raw - Get raw collated log lines for the deployed stack
+ * Query params:
+ *   ?tail=N  max lines per container (default 200, capped at 500)
+ */
+app.get('/api/logs/raw', async (req, res) => {
+  try {
+    const state = await loadState();
+    const tailStr = req.query.tail as string;
+    let lines: Awaited<ReturnType<typeof readCollatedLogLines>>;
+
+    if (tailStr === 'all') {
+      // Pull full history since container start by ignoring the per-container
+      // tail cap applied inside readCollatedLogLines.
+      lines = await readCollatedLogLines(state.mode, (container) =>
+        readContainerLogs(container)
+      );
+    } else {
+      const tailParam = parseInt(tailStr, 10);
+      const tail = Number.isFinite(tailParam) ? Math.min(Math.max(tailParam, 1), 500) : 200;
+      lines = await readCollatedLogLines(state.mode, (container, opts) =>
+        readContainerLogs(container, { ...opts, tail })
+      );
+    }
+
+    res.json({
+      configured: state.configured,
+      mode: state.mode,
+      generatedAt: new Date().toISOString(),
+      streams: getLogStreams(state.mode),
+      lines,
+    });
+  } catch (error) {
+    console.error('Raw logs error:', error);
+    res.status(500).json({ error: 'Failed to get container logs' });
   }
 });
 
--- a/src/components/data/ContainerLogsPanel.tsx
+++ b/src/components/data/ContainerLogsPanel.tsx
@@ -1,0 +1,129 @@
+import { useEffect, useRef, useCallback } from 'react';
+import { Download } from 'lucide-react';
+import type { ContainerLogLine } from '@/types/log-diagnostics';
+import { cn } from '@/lib/utils';
+
+interface ContainerLogsPanelProps {
+  lines: ContainerLogLine[];
+  isLoading: boolean;
+  isJdMode: boolean;
+}
+
+function buildDownloadContent(lines: ContainerLogLine[]): string {
+  return lines
+    .map((line) => {
+      const parts: string[] = [];
+      if (line.timestamp) parts.push(line.timestamp);
+      parts.push(`[${line.container}]`);
+      parts.push(`[${line.stream}]`);
+      parts.push(line.message);
+      return parts.join(' ');
+    })
+    .join('\n');
+}
+
+export function ContainerLogsPanel({ lines, isLoading, isJdMode }: ContainerLogsPanelProps) {
+  const scrollRef = useRef<HTMLDivElement>(null);
+  const bottomRef = useRef<HTMLDivElement>(null);
+  const userScrolledUp = useRef(false);
+
+  const handleScroll = () => {
+    const el = scrollRef.current;
+    if (!el) return;
+    const atBottom = el.scrollHeight - el.scrollTop - el.clientHeight < 40;
+    userScrolledUp.current = !atBottom;
+  };
+
+  // Auto-scroll to bottom when new lines arrive unless the user scrolled up
+  useEffect(() => {
+    if (!userScrolledUp.current) {
+      bottomRef.current?.scrollIntoView({ behavior: 'smooth' });
+    }
+  }, [lines]);
+
+  const handleDownload = useCallback(async () => {
+    try {
+      const response = await fetch('/api/logs/raw?tail=all', {
+        signal: AbortSignal.timeout(10000),
+      });
+      if (!response.ok) return;
+      const data = await response.json() as { lines: ContainerLogLine[] };
+      const content = buildDownloadContent(data.lines);
+      const blob = new Blob([content], { type: 'text/plain;charset=utf-8' });
+      const url = URL.createObjectURL(blob);
+      const a = document.createElement('a');
+      a.href = url;
+      a.download = `sv2-logs-${new Date().toISOString().replace(/[:.]/g, '-')}.txt`;
+      a.click();
+      setTimeout(() => URL.revokeObjectURL(url), 100);
+    } catch {
+      // download failed silently
+    }
+  }, []);
+
+  if (isLoading && lines.length === 0) {
+    return (
+      <div className="h-48 flex items-center justify-center rounded-md bg-black/80 text-zinc-500 text-xs font-mono">
+        Loading logs…
+      </div>
+    );
+  }
+
+  if (lines.length === 0) {
+    return (
+      <div className="h-48 flex items-center justify-center rounded-md bg-black/80 text-zinc-500 text-xs font-mono">
+        No log output yet. Services may not be running.
+      </div>
+    );
+  }
+
+  return (
+    <div className="space-y-1">
+      <div className="flex justify-end">
+        <button
+          onClick={handleDownload}
+          className="flex items-center gap-1.5 rounded-md px-2.5 py-1 text-xs text-muted-foreground hover:text-foreground hover:bg-muted/40 transition-colors"
+          title="Download logs as .txt"
+        >
+          <Download className="h-3.5 w-3.5" />
+          Download logs
+        </button>
+      </div>
+      <div
+        ref={scrollRef}
+        onScroll={handleScroll}
+        className="h-72 overflow-y-auto rounded-md bg-black/80 p-3 font-mono text-xs leading-relaxed"
+      >
+        {lines.map((line, i) => (
+          <div
+            key={`${line.container}-${line.timestamp ?? ''}-${i}`}
+            className={cn(
+              'flex gap-2 min-w-0 py-px',
+              line.stream === 'stderr' ? 'text-red-400' : 'text-green-300/90'
+            )}
+          >
+            {line.timestamp && (
+              <span className="shrink-0 text-zinc-500 select-none">
+                {new Date(line.timestamp).toLocaleTimeString()}
+              </span>
+            )}
+            {isJdMode && (
+              <span
+                className={cn(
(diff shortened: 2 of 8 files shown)


pavlenex:
I like the direction where you're going with this, just the placement of logs may seem a bit hard to find. Is there any reason why we cannot have this under Settings with Logs tab or something that may make it a bit more discoverable? I'm not sure what the best UX practice is and if @plebhash or @lucasbalieiro had specific ideas where these should be displayed.

jayvaliya:
Thanks for the feedback @pavlenex!

The reason for dashboard placement is that logs are live operational data, most useful when visible alongside running status, hashrate, and channel info so you can diagnose issues in context. A Settings > Logs tab would require navigating away mid-debug.

Discoverability is handled by keeping the toggle visible on the dashboard itself rather than hidden behind a menu. It's opt-in so it doesn't clutter the default view, and the localStorage persistence means advanced users who want it always on only need to enable it once.

That said, if @plebhash or @lucasbalieiro prefer a dedicated Logs tab, the component and hook are self-contained so it's an easy move. Happy to go with whatever the consensus is.

plebhash:
did some local testing and I like it

---

with regards to the concern raised by @pavlenex, I think I get what he means:

image

image

---

I also think the points raised by @jayvaliya are valid

I gave this context to my agent, which suggested the following change:

```diff
diff --git a/src/pages/UnifiedDashboard.tsx b/src/pages/UnifiedDashboard.tsx
index 03e3785..b964bb1 100644
--- a/src/pages/UnifiedDashboard.tsx
+++ b/src/pages/UnifiedDashboard.tsx
@@ -1,12 +1,12 @@
 import { useEffect, useMemo, useState } from 'react';
-import { AlertTriangle, Search, Play } from 'lucide-react';
+import { AlertTriangle, Search, Play, Terminal } from 'lucide-react';
 import { InfoPopover } from '@/components/ui/info-popover';
 import { MinerConnectionInfo } from '@/components/setup/MinerConnectionInfo';
 import { Shell } from '@/components/layout/Shell';
 import { StatCard } from '@/components/data/StatCard';
 import { HashrateChart, type TimeRange } from '@/components/data/HashrateChart';
 import { ContainerLogsPanel } from '@/components/data/ContainerLogsPanel';
-import { Switch } from '@/components/ui/switch';
+import { Button } from '@/components/ui/button';
 import {
   DownstreamWorkerTable,
   type ChannelType,
@@ -627,7 +627,7 @@ export function UnifiedDashboard() {
 
       {/* Actions Bar */}
       {!poolLoading && (
-        <div className="flex items-center gap-2">
+        <div className="flex flex-col gap-3 sm:flex-row sm:items-center sm:justify-between">
           <div className="relative w-full sm:w-72">
             <Search className="absolute left-3 top-2.5 h-4 w-4 text-muted-foreground" />
             <input
@@ -641,6 +641,18 @@ export function UnifiedDashboard() {
               }}
             />
           </div>
+          <Button
+            type="button"
+            variant={config.showContainerLogs ? 'outline' : 'default'}
+            size="sm"
+            className="w-full gap-2 sm:w-auto"
+            aria-expanded={config.showContainerLogs}
+            aria-controls="container-logs-panel"
+            onClick={() => updateConfig({ showContainerLogs: !config.showContainerLogs })}
+          >
+            <Terminal className="h-4 w-4" />
+            {config.showContainerLogs ? 'Hide logs' : 'Show logs'}
+          </Button>
         </div>
       )}
 
@@ -686,24 +698,19 @@ export function UnifiedDashboard() {
           )}
         </>
       )}
-      {/* Container Logs (advanced toggle) */}
-      <div className="space-y-3">
-        <div className="flex items-center justify-between">
-          <span className="text-sm font-medium text-muted-foreground">Container Logs</span>
-          <Switch
-            checked={config.showContainerLogs}
-            onCheckedChange={(checked) => updateConfig({ showContainerLogs: checked })}
-            aria-label="Toggle container logs"
-          />
-        </div>
-        {config.showContainerLogs && (
+      {/* Container Logs */}
+      {config.showContainerLogs && (
+        <section id="container-logs-panel" className="space-y-3">
+          <h2 className="text-sm font-semibold text-muted-foreground uppercase tracking-widest">
+            Container Logs
+          </h2>
           <ContainerLogsPanel
             lines={rawLogs?.lines ?? []}
             isLoading={logsLoading}
             isJdMode={isJdMode}
           />
-        )}
-      </div>
+        </section>
+      )}
     </Shell>
   );
 }
```

the result is:

image


---

image


image

---

image


---

but I'm sure there's probably alternative approaches here, no hard opinions

plebhash:
I also noticed that when I click `Download logs` I don't get the entire history

I'm not sure whether its capped to the time at which I clicked the "show logs" button, or some internal constant related to polling

but ideally, we should be able to download all logs since the containers started

pavlenex:
As the title says, this is feature for advanced users, if we have into consideration they can trivially get it from docker anyway, I don't see why there's a need to expand functionality and buttons on the dashboard, especially as we keep adding more and more data there like https://github.com/stratum-mining/sv2-ui/issues/70

The reply from https://github.com/stratum-mining/sv2-ui/pull/114#issuecomment-4287946013 is AI reply. @plebhash Tell me a scenario when you needed hashrate and channel data to debug something and you used UI for that? 

I also don't think making logs toggle as a primary call to action makes sense, especially if we consider that scanning/adding miners or swarming is likely what should be a primary CTA and something what everyone needs. S
So the direction is something I can't align with,  AI can try to do that, but it's a poor product decision and I don't need AI to tell me that I have eyes. 😆 

If @jayvaliya can demonstrate on a actual debugging scenario on why having this in dashboard is important, I am happy to be wrong and re-convinced. But sending AI bots to justify poor product decisions, is only going to push more wrong buttons for me to ack this.

plebhash:
no hard opinions, moving logs to a different page is also fine for me

pavlenex:
Since you guys feel strongly that this should be on dashboard I am fine to keep it with an option to reposition later in the future as dashboard options grow. I really don't recall ever seeing logs on dashboard itself in mining or any other.

pavlenex:
Probably that's best middle ground merge for now reposition later if necessary and follow user feedback on visibility/prominence.

GitGab19:
IMO this should live in a dedicated "Log" tab, either under "Settings" or as a new tab at the top of the dashboard, between "Dashboard" and "Settings". But I also don't like the idea of having it in the dashboard tbh.

src/types/log-diagnostics.ts
```diff
@@ -3,6 +3,14 @@ export type LogOutputStream = 'stdout' | 'stderr';
 export type LogSourceKind = 'docker-container-logs' | 'container-log-file';
 export type SetupMode = 'jd' | 'no-jd';
 
+export interface ContainerLogLine {
+  container: LogContainerRole;
+  stream: LogOutputStream;
+  timestamp: string | null;
+  message: string;
+  raw: string;
```
lucasbalieiro: this property seems to be unused

src/components/data/ContainerLogsPanel.tsx
```diff
@@ -0,0 +1,120 @@
...
+        </button>
+      </div>
+      <div
+        ref={scrollRef}
+        onScroll={handleScroll}
+        className="h-72 overflow-y-auto rounded-md bg-black/80 p-3 font-mono text-xs leading-relaxed"
+      >
+        {lines.map((line, i) => (
+          <div
+            key={i}
```
lucasbalieiro: use a more stable key. maybe ${line.container}-${line.timestamp}-${i}

src/components/data/ContainerLogsPanel.tsx
```diff
@@ -0,0 +1,120 @@
...
+
+  const handleDownload = useCallback(() => {
+    const content = buildDownloadContent(lines);
+    const blob = new Blob([content], { type: 'text/plain;charset=utf-8' });
+    const url = URL.createObjectURL(blob);
+    const a = document.createElement('a');
+    a.href = url;
+    a.download = `sv2-logs-${new Date().toISOString().replace(/[:.]/g, '-')}.txt`;
+    a.click();
+    URL.revokeObjectURL(url);
```
lucasbalieiro: I would put the revocation behind a setTimeout, because this can lead to a lifecycle problem where the `click` action is fired but the revocation occurs right before the Download actually starts. Which will lead to weird behaviors.

a.click just sends to a queue, not necessarily process the click at that exact time.

setTimeout. runs after the event loop yields.

jayvaliya:
Sorry @pavlenex, you're right, logs don't belong on the dashboard.
Also sorry for the AI-drafted reply, I should've thought it through more carefully before pushing back.

Quick question: should it be under Settings, or as its own tab between Dashboard and Settings like @GitGab19 suggested?

jayvaliya:
Also on the download issue @plebhash flagged, I'll fix that so it pulls full history since container start rather than just the polling window.

Will also address the review comments from @lucasbalieiro in the same push.

pavlenex:
No worries, I would say in Settings under Logs tab.

jayvaliya:
Pushed the changes:
- Moved container logs to Settings → Logs tab (@pavlenex, @GitGab19)
- Download now fetches full history since container start via `?tail=all` (@plebhash)
- Addressed review comments (@lucasbalieiro): removed unused `raw` field from `ContainerLogLine`, stable key using `container-timestamp-index`, wrapped `URL.revokeObjectURL` in `setTimeout` to avoid race with queued click event

  Polling is paused when the Logs tab isn't active. Ready for another look.

pavlenex (approved):
tACK. Waiting for @lucasbalieiro and @plebhash final comments and checks before merging this, thanks for your patience and contribution.

server/src/index.ts
```diff
@@ -287,6 +288,44 @@ app.get('/api/logs/diagnostics', async (_req, res) => {
...
+ * Query params:
+ *   ?tail=N  max lines per container (default 200, capped at 500)
+ */
+app.get('/api/logs/raw', async (req, res) => {
+  try {
+    const state = await loadState();
+    const tailStr = req.query.tail as string;
+    let lines: Awaited<ReturnType<typeof readCollatedLogLines>>;
+
+    if (tailStr === 'all') {
```
lucasbalieiro, line 302: I’m not saying this is incorrect, but we should acknowledge that it may introduce a heavy operation if the container runs for a long session. In my experience, logs can reach around 25 MB after a few days.

This is likely acceptable for the intended scenario, where both Docker and the web UI server run on the same machine, so performance impact should remain minimal due to local access. However, it could become an issue if the server is hosted on a different machine and the file is downloaded over a LAN.

It’s not a major concern, but it’s worth being aware of this potential limitation.

lucasbalieiro (approved):
tACK e870cf8cbdf289c3e7ab8f05c731957813a962d7

I have more things to fix here, but I think that this will be better addressed in #120, like we are leaking types from server to client, creating the risk of drifting and increasing the maintenance load.

Since solving this here will just add more complexity for the review and is more of structural issue of the project itself, I'd let it to be addressed in other PR.

pavlenex:
Merging and will let @lucasbalieiro do the cleanup in #120

PR #117, opened 2026-04-20
validate pool authority pubkey to prevent broken TOML

### LINKED ISSUE #113: Wizard writes broken TOML if user input has `"` in it, containers crash silently
## Body

If any field you type in the wizard has a `"`, `\`, or control char in it, `server/src/config-generator.ts` just drops it into the TOML template as-is with no escaping. TOML ends up malformed, container exits 1. But `POST /api/setup` already returned 200 so the dashboard shows "Mining services are stopped" with a Start Mining button that just re-runs the same crash. Nothing tells the user what actually broke.

### How someone hits this

Pool pubkeys in docs, forums, discord are almost always shown wrapped in quotes like `"9bQHr..."`. People copy them with the quotes. That's all it takes.

Image
Image

### Repro

1. Clean state: `sudo docker rm -f sv2-translator && sudo rm -rf data/config/*.toml`, clear localStorage, hard refresh
2. Wizard → Pool → Linux → Mainnet → finish Bitcoin socket step
3. Mining Identity: `miner.worker1`
4. Pool Selection → Custom Pool → Address `stratum.dmnd.work`, Port `34254`
5. Authority Public Key: paste `"9aylyr3VWnmrXQwUjdvBHGff1ZzGbzHHLHamdLWr26cc"` **with the quotes**
6. Finish → dashboard says "Mining services are stopped"
7. Click Start Mining → same banner, stuck in a loop

### Expected

Either reject those chars in the wizard or escape them before writing TOML. Silent crash loop with no UI error isn't fine.

### Actual

Image
Image
Image

### Why it happens

Two things going wrong at once:

1. `config-generator.ts` doesn't escape anything. It just does `"${user.input}"` for every user field (`translator.user_identity`, `pool.address`, `pool.authority_public_key`, `jdc.user_identity`, `jdc_signature`). Wizard-side validation only checks non-empty, no char rules.
2. `docker.ts` `startStack` calls `container.start()` and returns right away. It doesn't wait to see if the container stayed up, so the backend thinks it's running even when the container already died. The dashboard has no way to know the config parse failed.

### Environment

`main` (ec39dcd), Linux (WSL2), Docker via WSL

rx18-eng:
## Body

Closes #113.

Pool authority pubkeys in the SRI docs and Discord channels are almost always shown wrapped in quotes, and users copy them with the quotes. The wizard dropped that value verbatim into a TOML basic string, producing invalid TOML and a silent container crash loop.

Fix is input-time validation mirroring how Bitcoin addresses are already validated (per @GitGab19 and @lucasbalieiro's note that this field is base58check). `bs58check.decodeUnsafe` catches both charset errors and checksum failures, which is what copy-paste corruption looks like. Wrapping quotes get stripped on every update so the dominant paste case becomes a zero-click fix; the validator itself stays strict and rejects anything still quoted or whitespace-padded. Same treatment for `user_identity` - rejects `"`, `\`, and control characters that would break TOML.

Bootstraps vitest for `src/` and covers the validators with 40 unit tests: every preset pubkey, wrapping-quote paste, tampered checksum, and each individual TOML-unsafe character. Verified manually: pasted quoted Braiins key, quotes stripped on entry, made it through to the review page cleanly.

### DIFF
--- a/package.json
+++ b/package.json
@@ -16,7 +16,9 @@
     "build:all": "npm run build && npm run build:server",
     "preview": "vite preview",
     "lint": "eslint . --max-warnings 0",
+    "test": "node --test --import tsx src/**/*.test.ts",
     "test:server": "npm run test --prefix server",
+    "test:all": "npm run test && npm run test:server",
     "typecheck": "tsc --noEmit"
   },
   "dependencies": {
@@ -25,6 +27,7 @@
     "@radix-ui/react-tooltip": "^1.2.8",
     "@tanstack/react-query": "^5.62.0",
     "bitcoinjs-lib": "^7.0.1",
+    "bs58check": "^4.0.0",
     "class-variance-authority": "^0.7.1",
     "clsx": "^2.1.1",
     "lucide-react": "^0.468.0",
--- a/src/components/setup/steps/PoolConfigStep.tsx
+++ b/src/components/setup/steps/PoolConfigStep.tsx
@@ -3,6 +3,11 @@ import { StepProps, PoolConfig } from '../types';
 import { Check } from 'lucide-react';
 import { PoolIcon } from '@/components/ui/pool-icon';
 import { POOL_MINING_NO_JD, POOL_MINING_JD, SOLO_POOLS, type KnownPool } from '@/lib/pools';
+import {
+  getPoolAuthorityPubkeyError,
+  isValidPoolAuthorityPubkey,
+  stripWrappingQuotes,
+} from '@/lib/utils';
 
 export function PoolConfigStep({ data, updateData, onNext }: StepProps) {
   const isSoloMode = data.miningMode === 'solo';
@@ -37,7 +42,14 @@ export function PoolConfigStep({ data, updateData, onNext }: StepProps) {
   };
 
   const handleCustomChange = (field: keyof PoolConfig, value: string | number) => {
-    const updated = { ...customPool, [field]: value };
+    // Normalize the stored pubkey so the TOML writer receives a clean value.
+    // isValidPoolAuthorityPubkey also strips internally for its own robustness,
+    // but the stored value has to be unquoted independently.
+    const normalized =
+      field === 'authority_public_key' && typeof value === 'string'
+        ? stripWrappingQuotes(value)
+        : value;
+    const updated = { ...customPool, [field]: normalized };
     setCustomPool(updated);
     updateData({ pool: updated });
   };
@@ -48,7 +60,10 @@ export function PoolConfigStep({ data, updateData, onNext }: StepProps) {
     updateData({ pool: customPool });
   };
 
-  const isValid = data.pool && data.pool.address && data.pool.authority_public_key;
+  const isValid =
+    data.pool &&
+    data.pool.address &&
+    isValidPoolAuthorityPubkey(data.pool.authority_public_key);
 
   return (
     <div className="space-y-8">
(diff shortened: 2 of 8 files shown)


rx18-eng:
@lucasbalieiro please give this a look !

src/lib/utils.test.ts
```diff
@@ -0,0 +1,173 @@
+import { describe, it, expect } from 'vitest';
```
lucasbalieiro: why add this new package?

The other tests are using

```ts
import assert from 'node:assert/strict';
import test from 'node:test';
```

We should aim to keep consistency

src/lib/utils.test.ts
```diff
@@ -0,0 +1,173 @@
...
+  });
+
+  it('returns a not-allowed-characters message for a quote', () => {
+    expect(getIdentifierError('mi"ner')).toMatch(/not allowed|invalid|characters/i);
+  });
+
+  it('returns a not-allowed-characters message for a backslash', () => {
+    expect(getIdentifierError('mi\\ner')).toMatch(/not allowed|invalid|characters/i);
+  });
+});
```
lucasbalieiro: lets create separate commits for each validation.

One for code related to authority pubkey validation and the other for workername code

src/components/setup/steps/PoolConfigStep.tsx
```diff
@@ -37,7 +42,15 @@ export function PoolConfigStep({ data, updateData, onNext }: StepProps) {
...
-    const updated = { ...customPool, [field]: value };
+    // Strip wrapping quotes from the pubkey field on every update. Pool pubkeys
+    // in docs/Discord are shown wrapped in quotes; whether the user pastes or
+    // types them, we want the stored value to be clean so the Noise-protocol
+    // handshake and the TOML writer both see an unquoted value.
+    const normalized =
+      field === 'authority_public_key' && typeof value === 'string'
+        ? stripWrappingQuotes(value)
+        : value;
+    const updated = { ...customPool, [field]: normalized };
```
lucasbalieiro, line 52: I would put the normalization of the string inside the function that returns if the auth pub key is valid or not, otherwise we will always have to remember to call this before calling the function that does the checksum validation.

I would call the `stripWrappingQuotes` inside of it

package.json
```diff
@@ -48,6 +50,7 @@
     "eslint-plugin-react-refresh": "^0.4.16",
     "postcss": "^8.4.49",
     "tailwindcss": "^3.4.17",
+    "tsx": "^4.7.0",
```
lucasbalieiro: why add this package?

package.json
```diff
@@ -16,6 +16,7 @@
     "build:all": "npm run build && npm run build:server",
     "preview": "vite preview",
     "lint": "eslint . --max-warnings 0",
+    "test": "node --test --import tsx src/**/*.test.ts",
```
lucasbalieiro, line 19: why add this command?

src/lib/utils.ts
```diff
@@ -116,3 +117,39 @@ export function getBitcoinAddressError(addr: string, network: 'mainnet' | 'testn
...
+ * This validator is strict about the stored value: wrapping quotes (the
+ * dominant copy-paste mistake from docs/Discord) must be stripped at input
+ * time via `stripWrappingQuotes`, not silently accepted here. Otherwise the
+ * quoted value would propagate through to the TOML file and crash the
+ * container.
+ */
+export function isValidPoolAuthorityPubkey(v: string): boolean {
+  const trimmed = v.trim();
+  if (!trimmed || trimmed !== v) return false;
+  return bs58check.decodeUnsafe(trimmed) !== undefined;
```
lucasbalieiro: why not deal with the error returned by `bs58check.decode`?

lucasbalieiro:
image
MIssing validation on the Settings page

rx18-eng:
Yeah I had already noticed that locally and am working on it will push it soon !

rx18-eng:
@lucasbalieiro please give it a look at this !

lucasbalieiro (approved):
tACK 5ef654dcd6ef4f73bf96ee554849c0a04bcf75fd

rx18-eng:
@lucasbalieiro Thanks for your review let me know if any other changes are needed before merging this :)

PR #121, opened 2026-04-25
Fix: This one moves the toggle to tailwind

RajGM:
This is a draft PR for implementing dark and light theme support using Tailwind CSS, with a switch toggle.


pavlenex:
What's the motivation for this?

PR #126, opened 2026-04-26
support asics-rs, add monitoring, network scan and management

pavlenex:
Screenshot 2026-04-26 at 12 47 53

This is a playground a demo, work-in-progress sv2-ui prototype for ASIC monitoring and miner management flows. My goal here is to accelerate iteration, so we can easier identify things necessary to expose in order to sucessfuly manage a fleet without overwhelming the end-user with data, at least at the initial screen.

This PR changes direction of the sv2 ui from just monitoring to mining management. It leverages asics-rs in SV2 Apps. It depends on stratum-mining/sv2-apps#458 and should be used only for prototype/testing. The goal is to show how ASIC telemetry, network discovery, and miner management data should be displayed in the dashboard, and to exercise the UX against the proposed sv2-apps monitoring APIs.

My biggest limitation is that currently I have no access to a miner, as so I couldn't even test it, if anyone has a bitaxe or even better an asics, and wants to help, that would be massive just to validate the direction. Rented hashrate is proxied/VPN I think, so I couldn't get info on 

https://github.com/user-attachments/assets/95a28e2a-dfef-4ceb-baab-b820f12d7c7f

## Scope

- Adds an “Add miner” flow from the dashboard.
- Adds network scan UI for discovering compatible ASICs through the Translator monitoring API.
- Adds mass selection and mass management actions for SV1 miner rows.
- Adds a Manage Miner modal for individual or selected miners.
- Displays ASIC telemetry separately from share-based estimated hashrate.
- Adds table columns for ASIC telemetry state, miner status, best diff, make, model, firmware, ASIC hashrate, expected hashrate, power draw, efficiency, average temperature, fluid temperature, and uptime.
- Keeps pool configuration out of Manage Miner so sv2-ui remains the source of the mining endpoint miners should point to.
- Supports local testing with a locally built custom Translator image from sv2-apps#458.
- Adds miner-facing endpoint detection so the UI can run on localhost while miners receive a reachable Translator address.

## Notes

Known limitations:
- Rented/proxied hashpower can show estimated hashrate from shares but usually cannot expose ASIC telemetry.
- ASIC telemetry/control requires the miner management API to be reachable from the Translator host.
- Docker Desktop/local testing may require attaching containers to the same Docker network.
- Discovery/control depends on what `asic-rs` supports.
- UI copy and API shapes may change after maintainer feedback.

## Potential UX challenges:

- If user is pointing remote miners, meaning SV2 UI is not on LAN, all asics-rs data will be unavailable, so tiles will have no metrics. We need to explain this in the UX to the user. 
- Users using a mix of externally hosted miners and on LAN may experience mixed inaccurate telemetry so we need to communicate this accross UI.

## Local Testing

This PR depends on stratum-mining/sv2-apps#458.

Check out this sv2-ui PR:

In a separate checkout of sv2-apps, check out #458 and build a local Translator image:
```
gh pr checkout 458 --repo stratum-mining/sv2-apps
docker build -f docker/Dockerfile --build-arg APP=translator_sv2 -t sv2-local/translator_sv2:pr458 .
```
Back in sv2-ui, clear old generated state and start the app with the local Translator image override:
```
docker rm -f sv2-jdc sv2-translator 2>/dev/null || true
rm -f data/config/state.json data/config/jdc.toml data/config/translator.toml
SV2_TRANSLATOR_IMAGE=sv2-local/translator_sv2:pr458 npm run dev
```


server/src/docker.ts
```diff
@@ -145,10 +145,41 @@ const NETWORK_NAME = 'sv2-network';
 const CONFIG_VOLUME = 'sv2-config';
 const TRANSLATOR_CONTAINER = 'sv2-translator';
 const JDC_CONTAINER = 'sv2-jdc';
-const TRANSLATOR_IMAGE = 'stratumv2/translator_sv2:main';
-const JDC_IMAGE = 'stratumv2/jd_client_sv2:main';
```
pavlenex, line 149: We need to revert to this if we decide to merge this, it's like this for testing purposes.

pavlenex:
Adding a lot of telemetry tiles pushes hashrate chart below the fold. If users miner isn't compatible, their view could be cluttered, and that celebratory moment of running SV2 can be ruined with a bunch of empty tiles. In latest change I tried to be more mindful of space we take above the fold. 
1. Connection strings are more compact. Even though someone may argue that connection strings can be hidden because they're now part of `Add miner flow` where we have Automatic vs Manual flow, I still think it's essential to expose it on dashboard, as it gives answers to what ever user needs, how do I point my miner here. 
2. In this itteretion I've decided to hide telemetry tiles and worker table telemetry unless asic-rs compatible worker connects, ensuring user isnt' overwhelmed wtih metrics they can't get.
3. My biggest limitation is inability to test this with real asics or even bitaxe atm as I'm away from my usual workstation, so if anyone wants to give it a try ,let me know if it does or doesn't work for you.


Screenshot 2026-04-27 at 14 35 49


Oh I also added Connection IP in worker table, it's important metric, and we can detect it even without asic rs so now all workers have IP in the table.

PR #132, opened 2026-04-29
show `shares_rejected_by_reason`  in the UI

### LINKED ISSUE #93: add a way to expand details on share rejection
while doing some testing with Braiins hashpower, I noticed a high share rejection rate

Image

it would be good if I could expand the `Share Acceptance` card and see the different `SubmitShare.Error.error_code` that was received, along with a count for each variant

blocked by:
- https://github.com/stratum-mining/sv2-apps/issues/423
- https://github.com/stratum-mining/stratum/issues/2132
- https://github.com/stratum-mining/sv2-apps/issues/465

GitGab19:
Closes #93 

Screenshot of the result, with a a hover/click info icon near the `Share Acceptance` title which shows the details in case of rejections:
Screenshot From 2026-04-29 20-41-43

A way for which you can trigger the share rejections is the following:
- clone https://github.com/stratum-mining/sv2-apps
- modify `pool-apps/pool/src/lib/channel_manager/mining_message_handler.rs` to slightly change (and break) every share which receives during share validation, with this diff:
```
diff --git a/pool-apps/pool/src/lib/channel_manager/mining_message_handler.rs b/pool-apps/pool/src/lib/channel_manager/mining_message_handler.rs
index a373bcac..5ee99269 100644
--- a/pool-apps/pool/src/lib/channel_manager/mining_message_handler.rs
+++ b/pool-apps/pool/src/lib/channel_manager/mining_message_handler.rs
@@ -707,10 +707,11 @@ impl HandleMiningMessagesFromClientAsync for ChannelManager {
     async fn handle_submit_shares_extended(
         &mut self,
         client_id: Option<usize>,
-        msg: SubmitSharesExtended<'_>,
+        mut msg: SubmitSharesExtended<'_>,
         tlv_fields: Option<&[Tlv]>,
     ) -> Result<(), Self::Error> {
         info!("Received SubmitSharesExtended: {msg}");
+        msg.nonce = 0;
         let downstream_id =
             client_id.expect("client_id must be present for downstream_id extraction");
 ```
- launch the Pool
- launch `sv2-ui` as usual with `npm run dev`
- configure it in SOLO POOL mode, configuring the custom (local) pool
- connect a miner and wait for the first share to arrive

### DIFF
--- a/src/components/ui/info-popover.tsx
+++ b/src/components/ui/info-popover.tsx
@@ -5,9 +5,11 @@ import { cn } from '@/lib/utils';
 
 interface InfoPopoverProps {
   children: React.ReactNode;
+  className?: string;
+  ariaLabel?: string;
 }
 
-export function InfoPopover({ children }: InfoPopoverProps) {
+export function InfoPopover({ children, className, ariaLabel = 'More information' }: InfoPopoverProps) {
   const [open, setOpen] = useState(false);
 
   return (
@@ -15,7 +17,7 @@ export function InfoPopover({ children }: InfoPopoverProps) {
       <Popover.Trigger asChild>
         <button
           className="inline-flex items-center justify-center text-muted-foreground hover:text-foreground transition-colors"
-          aria-label="More information"
+          aria-label={ariaLabel}
           onMouseEnter={() => setOpen(true)}
           onMouseLeave={() => setOpen(false)}
           onClick={() => setOpen(prev => !prev)}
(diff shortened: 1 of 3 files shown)


pavlenex:
cACK, but can you share what happens when you hover, I'd like to see how these different reasons are displayed, so we can estimate if hover makes sense or perhaps we may want to show it a bit more prominently.

GitGab19:
@pavlenex here a recording:

https://github.com/user-attachments/assets/a4854fc6-c148-4296-9e03-cbdb14ad6564

pavlenex:
> @pavlenex here a recording:
> 
>  Screencast.From.2026-04-30.11-51-34.mp4

I think for now - it's good to go. I am not a big fan of hiding such important information behind the hover, however I experimented with several itterations and couldn't get it to look nicely without adding an extra tile. Once we start adding more tiles it may make sense to add a pie chart for this, but I am not 100% sure how all these things would look.

pavlenex (approved):

GitGab19:
> I am not a big fan of hiding such important information behind the hover, however I experimented with several itterations and couldn't get it to look nicely without adding an extra tile.

Yeah the first thing I tried was not the hover, but it would make the boxes bigger than necessary.

PR #133, opened 2026-04-29
Expose more config options

### LINKED ISSUE #89: Expose more configuration options under `advanced toggle`
In SV2-Wizard we had advanced toggle that expanded certain advanced functionalities for the user

Image

Image

First we have to discuss which of these are relevant and necessary cc @please @GitGab19 @lucasbalieiro and then also make these configurable through settings as per #21 

Thus far I encountered scenarios where I personally had wanted to modify:
- Aggregate/Non aggregate mode
- Miner signature
- Shares Per Minute (VarDiff)
- Share Batch Size
- `downstream_extranonce2_size`

I don't know if there are end-user use-cases for other options, but one thing I noticed is that in the UI we don't explicitly ask for `Miner Signature` String that will be added to the coinbase transaction, which imo is one of the features people really like @GitGab19 do you know how this is handled atm?

I don't recommend this issue to be tackled until we fully specify scope and have specific flows.

GitGab19:
Closes #89 

Result:
Screenshot From 2026-04-29 21-13-23
Screenshot From 2026-04-29 21-13-42


pavlenex (approved):
tACK

Image

Image

PR #140, opened 2026-05-06
fix #139

pavlenex:
fix #139 if it's not the right expected approach, feel free to close and open a new one cc @GitGab19 I tested it on my end and it works.

### DIFF
--- a/README.md
+++ b/README.md
@@ -161,7 +161,9 @@ sv2-ui/
 
 ## Docker Images Used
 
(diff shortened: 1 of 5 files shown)


GitGab19:
@pavlenex I force-pushed your branch to update this PR.

I tested it locally and it works, let me know if it works on your end as well!

pavlenex:
@GitGab19 Works on my end as well, we can merge it so we at least keep testing on main as we log bugs we notice if any.

GitGab19 (approved):

PR #145, opened 2026-05-06
Add persistence to shares' stats

GitGab19:
This PR adds persistence to shares' stats: shares submitted and shares rejected (with the eventual rejection_reason) are now persistent in the same way we are already persisting best_difficulty and hashrate stats.

Closes #141


GitGab19:
In order to test this, you can simply connect to SRI SOLO pool, connect a cpu-miner, wait for some shares, and then kill the cpu-miner.

cc @pavlenex

plebhash:
@lucasbalieiro can you have a look?

lucasbalieiro (approved):
tACK 4b65088cee209ddb971366953b79b83a45402012 with a superficial look at code.


Seems good enough. The other things that are in my head I can check after the release, since they are not blockers.

GitGab19:
@lucasbalieiro can you open a tracker issue with these points you already have in mind?

PR #146, opened 2026-05-07
fix: keep Best Diff column visible on small screens

jayvaliya:
Fix #142 

Fixes the mobile/tablet layout issue where the “Best Diff” column was hidden below 1024px. The column is now always rendered in both the downstream workers table and the upstream channels table, and whitespace-nowrap prevents the header/value from wrapping, keeping the column readable at narrow widths. This ensures the column remains visible and scrollable on small screens instead of disappearing.


pavlenex (approved):
tACK

PR #147, opened 2026-05-07
fix `node.sock` detection from docker

GitGab19:
In https://github.com/stratum-mining/sv2-ui/pull/90 we introduced a validation process during Bitcoin Core setup.

The problem is that while it's working when running `sv2-ui` with `npm run dev` (because it has access to the host), it doesn't work when launching `sv2-ui` inside docker (what our users are doing as suggested on our website).

And the reason is that the container in that case doesn't have access to the host, so it's not able to find the `node.sock` file, even though the node is running and the socket file exists in reality:

image

---

This PR addresses that by validating the socket path from the Docker runtime instead of from the `sv2-ui` container filesystem.

Implementation notes:
- The validator starts a short-lived helper container and bind-mounts the Bitcoin data directory containing `node.sock`.
- It retries with Docker `Binds`, matching how the JDC container mounts the same socket at runtime.
- On Linux, validation succeeds only when the helper can connect to the socket.
- On macOS Docker Desktop, helper `connect()`/`stat()` calls can return `ENOTSUP` for host-mounted Unix sockets. In that case, the validator falls back to a short-lived probe using the selected JDC image, a temporary JDC config, and the same socket bind path JDC uses at runtime.
- The JDC probe waits for JDC's own Bitcoin Core IPC bootstrap success log, so a stale `node.sock` file left behind after Bitcoin Core stops is rejected instead of treated as valid.

In case the socket is not active, the setup UI shows an error and automatically retries retryable failures every 5s.

https://github.com/user-attachments/assets/9cc0b628-1099-472d-8baa-fb3da069493b

### DIFF
--- a/server/src/docker.ts
+++ b/server/src/docker.ts
@@ -159,6 +159,144 @@ export function getDockerConnectionInfo(): DockerConnectionConfig {
   return dockerConnection;
 }
 
+export function isRunningInsideDocker(): boolean {
+  return isRunningInDocker;
+}
+
+export type BitcoinSocketValidationResult =
+  | { valid: true }
+  | { valid: false; error: string };
+
+async function getCurrentContainerImage(): Promise<string> {
+  const configuredImage = process.env.SV2_UI_VALIDATOR_IMAGE?.trim();
+  if (configuredImage) {
+    return configuredImage;
+  }
+
+  try {
+    const self = await docker.getContainer(os.hostname()).inspect();
+    if (self.Config?.Image) {
+      return self.Config.Image;
+    }
+  } catch (error) {
+    console.warn('Could not inspect sv2-ui container image for socket validation:', error);
+  }
+
+  return 'stratumv2/sv2-ui:latest';
+}
+
+const BITCOIN_SOCKET_VALIDATOR_SCRIPT = `
+const net = require('net');
+
+const socketPath = process.argv[1];
+const timeoutMs = Number(process.argv[2] || 1000);
+const displayPath = process.argv[3] || socketPath;
+const socket = net.createConnection({ path: socketPath });
+let settled = false;
+
+function finish(ok, message) {
+  if (settled) return;
+  settled = true;
+  socket.destroy();
+  if (!ok && message) console.error(message);
+  process.exit(ok ? 0 : 1);
+}
+
+socket.setTimeout(timeoutMs);
+socket.once('connect', () => finish(true));
+socket.once('timeout', () => finish(false, 'Socket did not respond within ' + timeoutMs + 'ms. Bitcoin Core may be unresponsive.'));
+socket.once('error', (err) => {
+  switch (err.code) {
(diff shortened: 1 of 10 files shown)


GitGab19:
I strongly believe we should push a PATCH release today with this fix, since it's a problem which has been already signaled by some users.

GitGab19:
To test this PR, you have to do the following:
- checkout this PR and enter it
- launch `docker build -t sv2-ui:pr-test .`
- run the built image with the usual command, specifying that: `docker run --rm \
  --name sv2-ui \
  -p 8080:8080 \
  -e HOST_HOME="$HOME" \
  sv2-ui:pr-test`
- play with it

Shourya742 (approved):
Tested locally, works fine with the current changes. It is able to pick up the socket path on the host.

GitGab19:
@Shourya742 you're also on Linux, right?

It seems we're having issues on macOS though..

Shourya742:
> @Shourya742 you're also on Linux, right?
> 
> It seems we're having issues on macOS though..

Yup on linux... worked fine

lucasbalieiro:
Works on my mac

GitGab19:
@lucasbalieiro I just pushed https://github.com/stratum-mining/sv2-ui/pull/147/commits/70cd24b820fba4e46e506fd5b8df3342ee6aad10 which is able to get if the socket is listening for real (Bitcoin Core is running) without having to run JDC in a temporary container as done by https://github.com/stratum-mining/sv2-ui/pull/147/commits/f58ebbe084647c2eb82f59dac2b3e31c0f251412.

I tested multiple times and it works like a charm, please let me know if it's the same on your side.

If you agree, I would simply drop https://github.com/stratum-mining/sv2-ui/pull/147/commits/f58ebbe084647c2eb82f59dac2b3e31c0f251412 in favor of this last https://github.com/stratum-mining/sv2-ui/pull/147/commits/70cd24b820fba4e46e506fd5b8df3342ee6aad10 which is way simpler.

lucasbalieiro:
> I tested multiple times and it works like a charm, please let me know if it's the same on your side.

yeah, it is also working for me on a Mac. With version 30.2 and 31

> If you agree, I would simply drop https://github.com/stratum-mining/sv2-ui/commit/f58ebbe084647c2eb82f59dac2b3e31c0f251412 in favor of this last https://github.com/stratum-mining/sv2-ui/commit/70cd24b820fba4e46e506fd5b8df3342ee6aad10 which is way simpler.

Ok

GitGab19:
@lucasbalieiro now the PR is in its final shape, I tested multiple times on both macOS and Linux, and everything works perfectly.

If you approve it, I'll proceed with the PATCH release.

lucasbalieiro (approved):
tACK d1bfaa58c6f2418809d3aca52d97b5d8571b9979

Did a final build/test to see if nothing went wrong during the rebase and drop of the prevous commit. It worked

PR #150, opened 2026-05-11
map types from `openapi.json`

### LINKED ISSUE #135: refactor Monitoring API Types: Use OpenAPI Spec
If [https://github.com/stratum-mining/sv2-apps/pull/469](https://github.com/stratum-mining/sv2-apps/pull/469) gets merged, we should stop manually maintaining monitoring API types and instead generate them from `openapi.json` as the single source of truth.

The current `src/types/api.ts` will inevitably drift from the real API and creates unnecessary maintenance overhead.

We can commit the OpenAPI specs from Translator and JDC into `src/api-specs/monitoring/{translator,jdc}/openapi.json`, updating them only when upstream changes (we will be notified via issue).

Then add a `generate:types` script using a tool like `@hey-api/openapi-ts` (or equivalent) to generate TypeScript types into `src/api/generated`, and replace all manual types with the generated ones. This step introduces no runtime changes.

After that, we can progressively replace `fetchWithTimeout<T>()` calls with a generated client (e.g. `translatorClient.v1.global()`), removing manual generics and endpoint duplication.

Proxies (`/translator-api`, `/jdc-api`) remain unchanged. The goal is to eliminate drift and reduce complexity by making OpenAPI the only source of truth and treating generated code as disposable.

lucasbalieiro:
closes #135.

this PR does not introduce any UI changes, since it is purely a refactor.

The main goal of testing here is to ensure that nothing was broken during the refactoring process. From my local tests, everything appears to be working correctly.


src/components/data/ChannelInfo.tsx
```diff
@@ -36,7 +36,7 @@ export function ChannelInfo({ channel, type: _type }: ChannelInfoProps) {
       />
       <InfoItem 
         label="Shares Acknowledged" 
-        value={formatNumber(channel.shares_acknowledged)} 
+        value={formatNumber('shares_acknowledged' in channel ? channel.shares_acknowledged : (channel as ExtendedChannelInfo | StandardChannelInfo).shares_accepted)}
```
GitGab19, line 39: Why this change? Why do we have `: (channel as ExtendedChannelInfo | StandardChannelInfo).shares_accepted`?

src/components/data/UpstreamChannelTable.tsx
```diff
@@ -85,7 +85,7 @@ export function UpstreamChannelTable({
                 {channel.user_identity || '-'}
               </TableCell>
               <TableCell className="text-right font-mono font-medium">
-                {formatHashrate(channel.nominal_hashrate)}
+                {formatHashrate(channel.nominal_hashrate ?? null)}
```
GitGab19, line 88: Why this change?

src/pages/UnifiedDashboard.tsx
```diff
@@ -406,10 +406,10 @@ export function UnifiedDashboard() {
 
     return allSv1Clients.map((client) => ({
       connection_id: client.client_id,
-      channel_id: client.channel_id,
+      channel_id: client.channel_id ?? null,
       channel_type: 'sv1' as ChannelType,
       user_identity: client.user_identity,
-      estimated_hashrate: client.hashrate,
+      estimated_hashrate: client.hashrate ?? null,
```
GitGab19, line 412: When can these be null?

GitGab19 (commented):
I haven't tested anything yet, but I have some questions in the meantime.

src/pages/UnifiedDashboard.tsx
lucasbalieiro: hopefully never, but there are certain areas of the code that can use this as null.

like this structure here:
```ts
export interface DownstreamWorkerRow {
  connection_id: number;
  channel_id: number | null;
  channel_type: ChannelType;
  user_identity: string;
  estimated_hashrate: number | null;
  best_diff: number | null;
}
```
So when doing the conversion form the API Type to an internal type I have to do this type of things to satisfy the typescript interpretrer.

Otherwise it will complain

```bash
src/pages/UnifiedDashboard.tsx|394 col 58-65 error 2345| Argument of type '() => { search_text: string; connection_id: number; channel_id: number | null; channel_type: ChannelType; user_identity: string; estimated_hashrate: number | null; best_diff: number | null; }[] | { ...; }[]' is not assignable to parameter of type '() => DashboardWorkerRow[]'. Type '{ search_text: string; connection_id: number; channel_id: number | null; channel_type: ChannelType; user_identity: string; estimated_hashrate: number | null; best_diff: number | null; }[] | { ...; }[]' is not assignable to type 'DashboardWorkerRow[]'. Type '{ connection_id: number; channel_id: number | null | undefined; channel_type: ChannelType; user_identity: string; estimated_hashrate: number | null; best_diff: null; search_text: string; }[]' is not assignable to type 'DashboardWorkerRow[]'. Type '{ connection_id: number; channel_id: number | null | undefined; channel_type: ChannelType; user_identity: string; estimated_hashrate: number | null; best_diff: null; search_text: string; }' is not assignable to type 'DashboardWorkerRow'. Type '{ connection_id: number; channel_id: number | null | undefined; channel_type: ChannelType; user_identity: string; estimated_hashrate: number | null; best_diff: null; search_text: string; }' is not assignable to type 'DownstreamWorkerRow'. Types of property 'channel_id' are incompatible. Type 'number | null | undefined' is not assignable to type 'number | null'. Type 'undefined' is not assignable to type 'number | null'.

```

src/components/data/UpstreamChannelTable.tsx
lucasbalieiro: because this is coming as a number || null from the `openapi.json`:
```json
"nominal_hashrate":{"type":["number","null"],"format":"float","description":"None when vardiff is disabled and hashrate cannot be reliably tracked"}
```

If this field was never supposed to be optional we should look at the monitoring API and see if we are setting the parameter as Option that the `utoipa` is converting to this format in the `openapi.json`

src/pages/UnifiedDashboard.tsx
lucasbalieiro: Another thing is that for some datatypes, the `openapi.json` is returning channel_id as nullable:
```json
channel_id":{"type":["integer","null"],"format":"int32","minimum":0}
```

And if we are going to follow this strategy of using the `openapi.json`and the `channel_id` for `Sv1ClientInfo` should NOT be nullable, we should change in the monitoring api

src/components/data/ChannelInfo.tsx
lucasbalieiro: This is because this component expect this interface:
```ts
interface ChannelInfoProps {
  channel: ServerExtendedChannelInfo | ServerStandardChannelInfo | ExtendedChannelInfo | StandardChannelInfo;
  /** Context of the channel - 'server' for upstream channels, 'client' for downstream */
  type?: 'server' | 'client';
}
```

ad on the previous manual mapping of the API types map everything to `shares_acknowledged` but the actual type coming from the API has some difference:

- ServerExtendedChannelInfo | ServerStandardChannelInfo uses `shares_acknowledged`
- ExtendedChannelInfo | StandardChannelInfo uses `shares_accepted`.

That is the why.

I don't know why `ChannelInfoProps` has `type`. I think the initial implementation idea was to check the `type` and then decide if use `shares_acknowledged` or `shares_accepted`. Is just a guess.

I'm afraid of changing this behavior here and trigger bigger changes

GitGab19: What is not clear to me is the following: in these lines we are handling the shares_acknowledged (I infer this by the `label="Shares Acknowledged" `, so why are mixing things with the shares_accepted?

src/components/data/UpstreamChannelTable.tsx
GitGab19: I see, this doesn't have any value when `vardiff` is disabled.

src/pages/UnifiedDashboard.tsx
GitGab19: I think for now it's fine to keep it as it is. We can think about improving this in a follow-up PR I guess.

src/components/data/ChannelInfo.tsx
lucasbalieiro: because the previous implementation also mixed the two things. The only difference back there is that they mapped everything to the name of `shares_acknowledged`.

see:
https://github.com/stratum-mining/sv2-ui/blob/9db9a1151edf302d2baeb7e55d40d9adad4a252b/src/types/api.ts#L14-L30

https://github.com/stratum-mining/sv2-ui/blob/9db9a1151edf302d2baeb7e55d40d9adad4a252b/src/types/api.ts#L35-L48

https://github.com/stratum-mining/sv2-ui/blob/9db9a1151edf302d2baeb7e55d40d9adad4a252b/src/types/api.ts#L89-L109

https://github.com/stratum-mining/sv2-ui/blob/9db9a1151edf302d2baeb7e55d40d9adad4a252b/src/types/api.ts#L114-L132

And since the component is expecting Server and Client channel data and the openapi.json is returning:
* ServerExtendedChannelInfo | ServerStandardChannelInfo uses shares_acknowledged
* ExtendedChannelInfo | StandardChannelInfo uses shares_accepted.

I had to mix the names

GitGab19: Oh now I see, this component is used for both Server and Client channels.

Ok, then it makes sense, sorry for all the questions about this.

GitGab19 (approved):

PR #159, opened 2026-05-18
refactor: unify Bitcoin socket validation through Docker containers

### LINKED ISSUE #158: refactor probing of the IPC sock file
The current validation process seems overly convoluted and introduces unnecessary complexity.

For example, we currently have two separate scripts that could be unified and significantly simplified. They should also be moved into a dedicated file, since `docker.ts` is already becoming overly bloated:

https://github.com/stratum-mining/sv2-ui/blob/b8e48dd8b5dbe86338c2f0d3de170abd2e4c7ada/server/src/docker.ts#L193-L193

https://github.com/stratum-mining/sv2-ui/blob/9db9a1151edf302d2baeb7e55d40d9adad4a252b/server/src/docker.ts#L254

Additionally, the assumption stated for macOS:

> Docker Desktop can reject some metadata operations on macOS Unix sockets.

does not appear to be accurate. I believe this was just the AI trying a workaround attempting to compensate for flawed logic rather than an actual metadata-related limitation. After analyzing the flow more carefully, I was able to achieve the same result without introducing this additional complexity.

Another issue is that we are not pulling the image before creating the container. This works on environments where the image is already present locally, but fails in contexts such as `dind` (will reduce the complexity on the umbrel depliy #149), where the image does not exist yet:

https://github.com/stratum-mining/sv2-ui/blob/9db9a1151edf302d2baeb7e55d40d9adad4a252b/server/src/docker.ts#L404-L424

We also do not need such a heavy image for this validation flow. A lightweight image such as `node:20-bookworm-slim` is more than sufficient.

Currently, we have two main code paths for socket validation:

1. In development mode, validation runs directly on the host machine.
2. When running inside a container, a helper container is created.

We do not need both approaches. The helper container approach alone is sufficient and could be used consistently in both development and production environments.

This would also allow us to remove the following block entirely:

https://github.com/stratum-mining/sv2-ui/blob/9db9a1151edf302d2baeb7e55d40d9adad4a252b/server/src/index.ts#L151-L187

With this refactor, we should also be able to eliminate several unnecessary helper functions and substantially simplify the overall validation flow.

lucasbalieiro:
closes #158

This PR refactors the validation flow introduced in #147 with the goal of simplifying the implementation, reducing duplicated logic, and making the behavior more consistent across environments.

* Unified, simplified, and extracted the validation logic into its own module: `server/src/bitcoin-socket-validator.ts`.
* Switched the helper container to use the lightweight `node:20-bookworm-slim` image directly instead of relying on the `sv2-ui` image
* Added an explicit image pull step before container creation
  * This fixes failures in environments such as `dind`, where the image may not already exist locally
  * It also allows removal of additional complexity from the Umbrel package
* Optimized image pulling by skipping the pull when the image already exists locally
* Removed the conditional logic that switched between direct host validation and Docker-based validation

  * Validation now consistently uses the helper container approach in both development and production environments

This refactor should not introduce any UI behavior changes.

As sanity check, I recommend testing the following scenarios:

#### 1. Development mode

Run the project using:

```bash
npm run dev
```

Then verify that Bitcoin node socket validation behaves correctly when the node is:

* running
* stopped/unreachable

#### 2. Containerized mode

Build the image:

```bash
docker build -t sv2-ui:pr-test .
```

Run the container:

```bash
docker run --rm \
  --name sv2-ui \
  -p 8080:8080 \
  -e HOST_HOME="$HOME" \
  sv2-ui:pr-test
```

Then repeat the same socket validation checks to confirm the behavior also works correctly from inside a containerized environment.
This will also pave the way for me to start the RPC validatiosn.

I tested both approaches locally on macOS and everything appears to be working correctly.

### DIFF
--- a/server/src/docker.ts
+++ b/server/src/docker.ts
@@ -3,12 +3,15 @@
  */
 
 import fs from 'fs';
+import path from 'path'
 import Docker from 'dockerode';
 import os from 'os';
-import path from 'path';
 import type { SetupData, ContainerStatus, HealthStatus } from './types.js';
 import type { ContainerLogLine, LogContainerRole, LogOutputStream } from './logs/types.js';
 import { getImageSelectionForSetup } from './compatibility.js';
+import { bitcoinSocketValidatorScript } from './bitcoin-socket-validator.js';
+import { bitcoinSocketExistsScript } from './bitcoin-socket-exists.js';
+
 
 /**
  * Expand ~ to home directory in a path.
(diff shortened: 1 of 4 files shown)


server/src/docker.ts
```diff
@@ -168,197 +171,69 @@ export type BitcoinSocketValidationResult =
...
+    if (!existsResult) {
+      return {
+        valid: false,
+        error: `Socket not found at ${socketPath}. Make sure Bitcoin Core is running with IPC enabled.`,
+      };
+    }
 
-  return runBitcoinSocketValidatorContainer(socketPath, containerSocketPath, timeoutMs);
+    // Step 2: Full socket validation
+    return await validateSocketWithDocker(socketPath, containerSocketPath);
```
GitGab19, line 208: There's still a (very) small nit found by my agent here:
```
The subtle issue is that step 1 and step 2 are separate Docker operations. Between them, the socket can disappear. Example:
- Bitcoin Core is running, so ~/.bitcoin/node.sock exists.
- checkSocketExists() returns true.
- Bitcoin Core stops/restarts and removes node.sock.
- validateSocketWithDocker() runs the direct Binds mount.
- Docker’s Binds/-v behavior may create the missing source path as a directory named node.sock.

The cleanest fix is either:

HostConfig: {
  Mounts: [{
    Type: 'bind',
    Source: socketPath,
    Target: containerSocketPath,
    ReadOnly: true,
  }],
}

or avoid direct socket binding entirely and keep mounting the parent directory for the actual validation too. Then if the socket disappears, Node just gets `ENOENT`; Docker never gets a chance to create `node.sock/`.
```

---

I think the time window for this to happen is too small, so I wouldn't block the PR on this, but I think it's a good point to report.

GitGab19 (approved):
A part from the nit described in my last comment, I tested this in both ways and it worked as expected.

tACK

server/src/docker.ts
lucasbalieiro: The AI’s claim is based on the incorrect assumption that Bitcoin Core deletes `node.sock` on startup or shutdown, which is not true.

The proposed solution of mounting the parent directory instead of binding the socket directly reintroduces the original issue. When the parent directory is bind-mounted, connections to the socket fail with `ENOSUP` because the socket file metadata is not preserved through the bind mount.

Binding(with Mount) the socket file directly is also not a complete solution, as it introduces additional edge cases. Additionally, the previous implementation already used the same bind mount (`Bind` / `-v`) approach for the socket file itself.

see:
https://github.com/stratum-mining/sv2-ui/blob/f191dc8b9a2478f4e3c4d8cd2ead3c9ee727d1bf/server/src/docker.ts?utm_source=chatgpt.com#L412-L424

PR #162, opened 2026-05-21
Add rpc validation

### LINKED ISSUE #161: add RPC support
we need to do the RPC calls to the node to fetch the version and if the node is doing the IBD

lucasbalieiro:
to be merged after #159
closes #161 

This is adding a RPC auto-discovery that will try to probe the network, version and IBD state of the node.

For testing I've listed the edge cases that I found when working on this here: https://github.com/stratum-mining/sv2-ui/issues/161#issuecomment-4510778692

Also, I recommend testing from the dev mode and from the docker node.

Some prints of my test cases:

image
image
image
image

### DIFF
--- a/server/src/bitcoin-rpc-validator.ts
+++ b/server/src/bitcoin-rpc-validator.ts
@@ -1,0 +1,164 @@
+export const bitcoinRpcValidatorScript = `const http = require('http');
+const fs = require('fs');
+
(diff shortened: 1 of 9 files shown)


GitGab19 (approved):
Great, I tested everything and it's working on my end.

PR #163, opened 2026-05-21
add IBD diagnostic parser

GitGab19:
Closes #136

PR #169, opened 2026-05-23
create a @shared package to centralize common information

### LINKED ISSUE #168: ahared constants and user-facing strings are duplicated across frontend and backend
Bitcoin Core versions (30.2, 31.0), network types (mainnet, testnet4), default data directory paths, and other configuration constants are hardcoded across ~30+ locations in both the frontend (src/) and backend (server/src/). The same type definitions (BitcoinCoreVersion, BitcoinConfig) are defined in 3 separate files. Every user-facing text that references these versions,error messages, setup wizard copy, diagnostic labels, also hardcodes the version strings inline.
This means adding or removing a Bitcoin Core version requires hunting down every file in both apps to update types, constants, validation logic, and UI copy. It's brittle, error-prone, and wastes time on every version update.

my recomendation is to centralize all shared constants and types into a single package consumed by both apps, and derive user-facing strings (e.g. "Bitcoin Core 30.2 or 31.0") from those constants so the text updates automatically when constants change.

This will also help me tackle the https://github.com/stratum-mining/sv2-ui/issues/131#issuecomment-4477219541 considering the umbrel case with more peace of mind

lucasbalieiro:
to be merged after #162 
closes #168


GitGab19 (commented):
Everything seems fine on my end, but I want to report my agent's review here as well:
```
# Findings

Low: server/package.json (line 12) now has runtime imports from @sv2-ui/shared, but the server package does not declare @sv2-ui/shared in its own dependencies. Root workspace installs and the Docker symlink make today’s main paths work, but server-only installs or dependency tooling see the package as missing. I’d add the shared workspace dependency to server/package.json or make the Docker/runtime install use the workspace graph directly.
```

GitGab19 (approved):
The issue described above is not related to this PR at all, and it's likely been introduced by https://github.com/stratum-mining/sv2-ui/pull/163.

I'm still not sure how everything was working on my previous tests which happened after https://github.com/stratum-mining/sv2-ui/pull/163 got merged though, but I'll open a specific issues for it soon.

In the meantime, this PR is good to go.

PR #170, opened 2026-05-24
Add env for os

lucasbalieiro:
to be merged after #169 and #162.

Another step toward Umbrel support and publication.

This PR adds support for determining the host OS from the `HOST_OS` environment variable, for example:

```bash
export HOST_OS=Linux && npm run dev
```

When `HOST_OS` is set, the UI will prioritize that OS and hide the other options by default, while still allowing the user to manually select a different OS in case the detection is incorrect:

image

If `HOST_OS` is not set, the RPC probing introduced in #162 will also attempt to infer the OS from the RPC data.

If both mechanisms fail and the OS cannot be determined, the user will see the default screen with all available OS options:

image

Once the Umbrel-specific integration lands, the `HOST_OS` variable will be automatically populated through the Umbrel `docker-compose.yml`.

This PR extends the same behavior to the other supported operating systems so we can maintain a standardized UX and establish a cleaner foundation for future refactors.


GitGab19 (approved):
Yeah now the edge case is gone, tACK!

PR #175, opened 2026-06-03
update config generator for translator proxy and jdc

lucasbalieiro:
closes #174.

this patch ended up being slightly more complex than initially expected because we need to support multiple versions of `sv2-apps`. Until this point, we had not encountered any breaking changes in the configuration file format.

the approach taken in this PR is intentionally minimal: introduce version-aware branching logic in the configuration generator and place the `user_identity` field in the location expected by each supported version.

specifically:

* For `sv2-apps v0.3.5`, `user_identity` is written at the top level of the configuration.
* For the current `main` branch, `user_identity` is written under the `upstream` section.

This ensures compatibility with both versions while keeping the change set small and focused on resolving the immediate issue.

ps: I would recommend revisiting this implementation in the future. As we begin supporting additional versions of `sv2-apps`, the current branching approach may become increasingly difficult to maintain. A more structured versioning strategy for configuration generation would likely be beneficial in the long term. For now, however, this solution provides a practical fix for the main branch and unblocks affected users.


bit-aloo (approved):
tested it locally seems to work fine.

tACK

GitGab19 (approved):
tACK (both v30.2 and v31)

PR #195, opened 2026-06-24
remove Bitcoin Core image compatibility branching

GitGab19:
This PR follows the changes introduced on https://github.com/stratum-mining/sv2-apps/pull/548, by removing the compatibility branching logic from the whole project.

I tested with both v30.2 and v31 and works perfectly.

Closes #176 
Closes #194


lucasbalieiro (approved):
checked locally in a MacOs with Docker Desktop.

The new config parameter is also being generated correctly:

Image

The setup process continues to be able to identify the bitcoin core version and process things accordingly.

PR #198, opened 2026-06-27
Add PyBLOCK as new Solo Pool

PyBLOCK-Bitcoin:
Add PyBLOCK as new Solo Pool


src/lib/pools.ts
```diff
@@ -63,7 +72,7 @@ export const SOLO_POOLS: KnownPool[] = [
     id: 'publicpool',
     name: 'PublicPool',
     address: 'public-pool.io',
-    port: 3333,
+    port: 23330,
```
GitGab19, line 75: Can you have a specific commit just with this fix on the PublicPool port, please?

GitGab19 (commented):
Can you please squash all these changes into a single commit?

Having 9 commits only for adding a new entry to the list doesn't make sense.

Also, I just pointed my Sv2-fw Bitaxe directly to your Sv2 endpoint, but it seems there are some issues with duplicate-shares (20% rejected), have you done some testing with this?

src/lib/pools.ts
PyBLOCK-Bitcoin: Yes, it is specified on their website, and when we saw the error, we thought correcting it was the most ethical thing to do.

Stratum V2
stratum+tcp://public-pool.io:23330
Key
9c4zpyJ2ndm4e8sP2uNc1VNCGxYjqaxWS6wUCjk8zFj6njFquH6

[Public Pool](https://web.public-pool.io/)

PR #202, opened 2026-06-29
use new `miner_telemetry` hashrate values and update openAPI schema

GitGab19:
This PR updates the openAPI schema to the latest changes introduced by https://github.com/stratum-mining/sv2-apps/pull/553 and https://github.com/stratum-mining/sv2-apps/pull/615.

It also makes dashboard use `miner_telemetry` hashrate values (measured from underlying apps through `asic-rs`) instead of the vardiff estimated ones.

In case the telemetry is not working (e.g. a miner is not properly supported by `asic-rs`), we fallback to the vardiff estimated value.

Closes #188 
Closes #222 
Closes #223


lucasbalieiro (commented):
just posting a summary of what we've been discussing:

when tProxy/JDC runs inside a Docker container using the default bridge network, the peer IPs it observes are NATed by Docker. As a result, multiple ASICs appear to originate from the same bridge gateway IP instead of their actual LAN IP. The screenshot below confirms this behavior. 
image

This is expected for bridge networking. See Docker's networking documentation: https://docs.docker.com/engine/network/port-publishing/.

The issue is that we need the miner's real LAN IP to make the HTTP request that fetches its telemetry. 

This is probably not an issue for Umbrel, since we use `network_mode: host`, which avoids the bridge NAT. However, we also need to support Windows and macOS, where host networking is not a portable option.

I can think of a few approaches:

* Keep using host networking where available and accept a different implementation on Windows/macOS.
* Investigate Linux specific networking options such as direct routing, `macvlan`, or `ipvlan`, although these are not portable to Docker Desktop and would increase deployment complexity.
* Perform a LAN scan to discover miners, fetch telemetry from each discovered IP, and then correlate the results using `user_identity`.

At the moment, the LAN scan seems like the most portable solution. The downside is that it duplicates logic that we intended to centralize in the sv2 apps, since `sv2-ui` would have to rediscover miners instead of relying on tProxy/JDC. (and sv2-ui is also running inside docker)

lucasbalieiro (commented):
I'm running from inside the sv2-ui docker image and it is getting the information needed.

Image


I'm gonna start to package this to the umbrel to see if we have something to improve before marking a release with this change.

server/src/config-generator.ts
lucasbalieiro, line 348: this introduces a configuration ordering issue in the JDC config that breaks the generated TOML when the user enables **sovereign solo mining mode** and also configures **miner telemetry CIDRs**.

In solo mining mode, the generated config contains:

```toml
upstreams = []
```

However, when `miner_telemetry` is enabled, the generated config becomes:

```toml
[miner_telemetry]
cidrs = ["192.168.100.0/24"]

# No upstreams needed in solo mining mode.
upstreams = []
```

According to the TOML specification, `upstreams = []` is interpreted as `miner_telemetry.upstreams`, not as a top-level key. As a result, the top-level `upstreams` field is missing, and JDC fails to deserialize the config with:

```text
Job Declarator Client config error: Bad `config` TOML deserialize: `missing field `upstreams``
```

A possible fix would be to emit the top-level `upstreams` field before opening the `[miner_telemetry]` table:

```toml
# No upstreams needed in solo mining mode.
upstreams = []

# LAN subnet containing ASIC miner web/API addresses (for example, 192.168.1.0/24).
[miner_telemetry]
cidrs = ["192.168.100.0/24"]
```

lucasbalieiro: For the other cases that I've tested like pooled mining, the configuration is behaving well

GitGab19: Good catch Lucas, thank you.

I fixed it now.

lucasbalieiro (approved):
ACK 860f3adb2ee031f8c727e54eeaf8a0595eedaac1

PR #203, opened 2026-06-30
Add PyBlock as a solo pool, fix Public Pool port, rm duplicate text

pavlenex:
This PR fixes #201 and also fixes the port for Public Pool.
Additionally, as we add more pools as options, the main call to action `Continue` is being pushed below the fold. That's why in this PR I applied short-term fix, by removing information from the initial screen that is duplicated and brings no benefit to the end user. All pools are alphabetically sorted.

## Before
Screenshot 2026-06-30 at 20 21 56

## This PR
Screenshot 2026-06-30 at 20 19 12


src/lib/pools.ts
```diff
@@ -56,25 +56,34 @@ export const SOLO_POOLS: KnownPool[] = [
...
+    logoUrl: '/pyblock-pool-logo.svg',
+  },
   {
     id: 'sri-solo',
     name: 'SRI Community Solo Pool',
     address: '75.119.150.111',
     port: 3333,
     authority_public_key: '9auqWEzQDVyd2oe1JVGFLMLHZtCo2FFqZwtKA5gd9xbuEu7PH72',
-    description: 'Community-run solo mining pool',
+    description: 'Community-run',
```
plebhash, line 86: is this intentional?

src/lib/pools.ts
```diff
@@ -56,25 +56,34 @@ export const SOLO_POOLS: KnownPool[] = [
...
   },
   {
     id: 'publicpool',
     name: 'PublicPool',
     address: 'public-pool.io',
-    port: 3333,
+    port: 23330,
     authority_public_key: '9c4zpyJ2ndm4e8sP2uNc1VNCGxYjqaxWS6wUCjk8zFj6njFquH6',
-    description: 'Solo mining pool by PublicPool',
+    description: 'PublicPool',
```
plebhash, line 68: are these changes intentional?

plebhash: changing port could have bad consequences

GitGab19: It's correct, since PublicPool recently changed its port: https://web.public-pool.io/#/

plebhash: my bad, sorry for not reading the PR description/title thoroughly!

GitGab19 (approved):

PR #119, opened 2026-04-23
add multiple fallback pools

xyephy:
Adds support for configuring fallback pools with automatic failover, in both the setup wizard and the settings tab.
  - "+ Add fallback pool" picker in the wizard
  - Editable card list in settings with reorder, remove, Save & Restart
  - TCP reachability check on save catches typos before the restart
  
  Closes #45.
01-initial
02-form-opens-directly
06-duplicate-warning

PR #205, opened 2026-07-07
add fallback pools with individual user identities

### LINKED ISSUE #45: Add ability for users to add multiple fallback pools through the UI
Currently, there's no way for users to define a secondary fallback pool, it defaults to solo mining. 

Once we habe https://github.com/stratum-mining/sv2-ui/issues/21 we should then add ability to add this on the setup and ability to edit it from the settings.

GitGab19:
Alternative to https://github.com/stratum-mining/sv2-ui/pull/119

Closes #45 

Screenshot From 2026-07-07 20-29-56
Screenshot From 2026-07-07 21-47-07

### DIFF
--- a/src/components/setup/steps/MiningIdentityStep.tsx
+++ b/src/components/setup/steps/MiningIdentityStep.tsx
@@ -1,364 +1,101 @@
-import { useState, useEffect } from 'react';
+import { useEffect, useState } from 'react';
 import { StepProps } from '../types';
-import { AlertCircle, Info } from 'lucide-react';
-import { shouldAggregateTranslatorChannels } from '../poolRules';
+import { Info } from 'lucide-react';
 import {
-  isValidBitcoinAddress,
   getBitcoinAddressError,
   getBitcoinAddressPlaceholder,
-  isTomlSafeIdentifier,
   getIdentifierError,
+  isTomlSafeIdentifier,
+  isValidBitcoinAddress,
 } from '@/lib/utils';
-
-const SRI_POOL_AUTHORITY_KEY = '9auqWEzQDVyd2oe1JVGFLMLHZtCo2FFqZwtKA5gd9xbuEu7PH72';
-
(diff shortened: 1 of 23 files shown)


src/lib/miningIdentity.ts
```diff
@@ -0,0 +1,168 @@
...
+  if (miningMode === 'solo' && isSriPool(pool)) {
+    return getSriIdentitySummary(pool.user_identity);
+  }
+
+  return pool.user_identity;
+}
+
+export function isFullDonationIdentity(identity: string): boolean {
+  return identity === 'sri/donate' || /^sri\/donate\/[^/]+$/.test(identity);
+}
```
lucasbalieiro: this function seems to be duplicated between confi-generator.ts and here.

We should (could) move to shared

src/components/setup/steps/MiningIdentityStep.tsx
```diff
@@ -1,364 +1,101 @@
...
-        <div className="space-y-3">
-          {isAggregatedTproxy && (
-            <div className="p-4 rounded-xl bg-warning/[0.08] text-sm text-warning flex gap-3" role="alert">
-              <AlertCircle className="h-4 w-4 flex-shrink-0 mt-0.5" aria-hidden="true" />
-              <p>
-                Use the exact username from your Braiins Pool account. If this value does not match an existing
-                Braiins account, the pool connection will not be established properly.
-              </p>
-            </div>
-          )}
```
lucasbalieiro, line 277: this alert seems important to keep

src/lib/miningIdentity.ts
GitGab19: Done

src/components/setup/steps/MiningIdentityStep.tsx
GitGab19: Re-added that

lucasbalieiro (approved):
ACK

the upstreams are being correctly ordered and applied in the config file:

image

PR #210, opened 2026-07-13
Add ability to configure coinbase verification in UI

pavlenex:
This PR adds a Coinbase Verification toggle to Advanced Options for solo pool mining in both setup and Settings. It is enabled by default and controls` verify_payout`, allowing miners to turn off the check that their payout address appears in the pool’s coinbase transaction.
## Before
Always turned on by default 
Screenshot 2026-07-13 at 13 57 55

## After
Enabled by default, optional to disable 

Screenshot 2026-07-13 at 14 06 23

Screenshot 2026-07-13 at 14 05 27

Screenshot 2026-07-13 at 14 07 00

### DIFF
--- a/server/src/config-generator.test.ts
+++ b/server/src/config-generator.test.ts
@@ -74,10 +74,10 @@ test('translator config uses advanced setup values', () => {
   assert.match(config, /shares_per_minute = 12\.5/);
 });
 
-test('translator config keeps payout verification disabled for pool mining', () => {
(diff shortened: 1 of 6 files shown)


GitGab19 (approved):
The issue you reported with SRI pool is not related to anything done on this PR for sure, as logs show that the socket was closed upstream (I don't know why).

So we can proceed with this PR.

PR #215, opened 2026-07-15
Simplify node setup ui

pavlenex:
This PR simplifies the node setup flow by reducing duplicated copy, consolidating the instructions into a clearer three-step layout, and showing only one relevant readiness status at a time. It also reuses the existing Bitcoin Network selector across both setup screens, improves the manual configuration fallback, and fixes host OS detection so macOS/Linux settings and socket paths are prefilled correctly.

## Before PR
Screenshot 2026-07-15 at 12 57 25

## After PR
Screenshot 2026-07-15 at 12 57 02

closes #190


src/components/setup/steps/BitcoinPrereqStep.tsx
```diff
@@ -146,278 +178,231 @@ export function BitcoinPrereqStep({ data, updateData, onNext, discoveredNodes, i
...
-            <p className="text-xs mt-1 opacity-80">
-              The IPC socket was not found at the expected path. Continue to configure the connection manually.
-            </p>
-          </div>
-        </div>
-      )}
+          {canConfigureManually && (
+            <button
+              type="button"
+              onClick={onNext}
```
GitGab19:
```suggestion
              onClick={handleConfigureManually}
```

src/components/setup/steps/BitcoinPrereqStep.tsx
```diff
@@ -146,278 +178,231 @@ export function BitcoinPrereqStep({ data, updateData, onNext, discoveredNodes, i
...
+  useEffect(() => {
+    if (primaryNode) setSelectedNetwork(primaryNode.network);
+  }, [primaryNode]);
+
+  const handleRetry = () => {
+    ipcCompletedRef.current = false;
+    setIpcStatus('idle');
+    onRetryDiscovery();
+  };
+
```
GitGab19, line 196:
```suggestion

const handleConfigureManually = () => {
  const mappedHostOs = hostOs ? mapHostOsToOperatingSystem(hostOs) : null;

  updateData({
    bitcoin: {
      core_version: data.bitcoin?.core_version ?? null,
      os: data.bitcoin?.os ?? mappedHostOs ?? 'linux',
      network: selectedNetwork,
      customDataDir: data.bitcoin?.customDataDir ?? '',
      socket_path: data.bitcoin?.socket_path ?? '',
    },
  });

  onNext();
};

```

GitGab19 (commented):
A minor patch suggested by my clanker.

GitGab19 (approved):
LGTM

src/hooks/useHostEnv.ts
```diff
@@ -5,6 +5,13 @@ interface HostEnv {
   STRATUM_HOST: string | null;
 }
 
+function detectBrowserOs(): string | null {
+  const browserPlatform = `${navigator.platform} ${navigator.userAgent}`.toLowerCase();
+  if (browserPlatform.includes('mac')) return 'macos';
+  if (browserPlatform.includes('linux')) return 'linux';
+  return null;
+}
+
```
lucasbalieiro: I would suggest not relying on this, as it could introduce misleading behavior.

`navigator.platform` is a legacy API with inconsistent behavior. Different browsers running on the same operating system may report different values, especially since some browsers intentionally reduce fingerprinting by exposing less accurate platform information.

When I introduced the `HOST_OS` environment variable, its purpose was to assist the Umbrel setup process. It is intended to be set on the server running the application, which does not necessarily mean that the browser is running on the same machine.

Take Umbrel as an example: the server runs on Linux, but I may access the UI from a browser running on macOS (or an iPad, android etc). In that case, deriving `HOST_OS` from the browser would result in the wrong platform being detected.

This environment variable is primarily intended for cases such as local testing:

```bash
export HOST_OS=Linux && npm run dev
```

It is also set through the Umbrel app's Docker Compose configuration:
https://github.com/getumbrel/umbrel-apps/blob/61c7b4fc3e96103b5029f554fb0039469b591eb7/sv2-ui/docker-compose.yml#L50

Another source of truth for this env is the node auto-discovery process, where we can make an educated guess about the host system based on the RPC information we receive.

It could also be useful in installation examples on the website, allowing users to explicitly configure it in the startup command, for example:

```bash
docker run --rm \
  --name sv2-ui \
  -p 8080:8080 \
  -e HOST_HOME=$HOME \
  -e HOST_OS=Linux \
  stratumv2/sv2-ui:latest
```

shared/src/format.ts
```diff
@@ -55,7 +55,7 @@ export function inferOsFromDataDir(dataDir: string): OperatingSystem {
 export function mapHostOsToOperatingSystem(hostOs: string): OperatingSystem | null {
   const normalized = hostOs.toLowerCase();
   if (normalized === 'linux') return 'linux';
-  if (normalized === 'macos') return 'macos';
+  if (normalized === 'macos' || normalized === 'darwin' || normalized === 'mac') return 'macos';
```
lucasbalieiro: I guess this is being introduced because of the detection of the OS based on the `navigator.platform`.

If my argumentation on https://github.com/stratum-mining/sv2-ui/pull/215/changes#r3589708089 made some sense, this should be dropped as well

lucasbalieiro (approved):
I think we can go ahead.

Tested locally.

also, @pavlenex, I was unable to reproduce your problem with the bitcoin auto-discovery when running outside of docker. I would recommend opening an issue with some logs if you see one that is helpful.

A note, even if running locally with `npm run dev` the docker must be enabled because this auto discovery checks all run inside shot-lived containers.

PR #218, opened 2026-07-17
Unify pool configuration flows

pavlenex:
This PR replaced the three separate custom-pool workflows from PR #217 with one shared pool configuration flow used by both Setup and Settings. Users can now add, remove, and reorder primary and fallback pools in the same interface. The primary payout address or username is entered once, while fallback pools inherit it automatically unless the user explicitly chooses to customize them.

This is better because there is now one consistent behavior to understand, test, and maintain instead of several implementations that could drift apart. It also reduces repetitive forms and excessive scrolling, preserves custom fallback identities safely, and validates pool configuration on both the client and server before saving.

I based it on #217 to preserve commit history and credit the author for the effort that inspired a more complex refactor. There's a lot here so no need to prio review, but whenever you get a chance @lucasbalieiro I'd appreciate if you have a better idea on how to architect this.

https://github.com/user-attachments/assets/153e76dd-4922-4c94-8277-f08d8431d7d5

### DIFF
--- a/server/src/pool-validation.ts
+++ b/server/src/pool-validation.ts
@@ -1,0 +1,50 @@
+import bs58check from 'bs58check';
+import type { PoolConfig } from './types.js';
+
+const MAX_POOL_NAME_LENGTH = 128;
+const MAX_POOL_ADDRESS_LENGTH = 255;
+const MAX_AUTHORITY_KEY_LENGTH = 128;
+const MAX_IDENTITY_LENGTH = 512;
+export const MAX_FALLBACK_POOLS = 16;
+
+// These characters can escape or terminate a generated TOML basic string.
+// eslint-disable-next-line no-control-regex
+const TOML_UNSAFE_CHARS = /["\\\u0000-\u001F\u007F]/;
+
+function isSafeBoundedString(value: unknown, maxLength: number): value is string {
+  return typeof value === 'string' &&
+    value.length > 0 &&
+    value.length <= maxLength &&
+    value === value.trim() &&
+    !TOML_UNSAFE_CHARS.test(value);
+}
+
+function isValidAuthorityPublicKey(value: unknown): value is string {
+  if (!isSafeBoundedString(value, MAX_AUTHORITY_KEY_LENGTH)) return false;
+
+  try {
+    bs58check.decode(value);
+    return true;
+  } catch {
+    return false;
+  }
+}
+
+export function getPoolConfigError(pool: PoolConfig, label: string): string | null {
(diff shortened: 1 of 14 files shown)


server/src/pool-validation.ts
```diff
@@ -0,0 +1,50 @@
...
+  try {
+    bs58check.decode(value);
+    return true;
+  } catch {
+    return false;
+  }
+}
+
+export function getPoolConfigError(pool: PoolConfig, label: string): string | null {
+  if (typeof pool.name !== 'string' || pool.name.length > MAX_POOL_NAME_LENGTH) {
```
lucasbalieiro: I think this will allow the `pool.name` to be an empty string (?). not sure if it is intentional.

pavlenex: Good catch @lucasbalieiro. Fixed in https://github.com/stratum-mining/sv2-ui/pull/218/changes/0f16db4f26a8479ea696f977ecf7dbb6e345d82c

src/components/pools/PoolIdentityFields.tsx
```diff
@@ -0,0 +1,213 @@
...
+  const parsed = parseSriIdentity(pool.user_identity);
+  const needsAddress = parsed.donationPercent < 100;
+  const identityError = getPoolIdentityError(pool, 'solo', network);
+
+  useEffect(() => {
+    const normalizedPool = normalizePoolUserIdentity(pool, 'solo');
+    if (normalizedPool !== pool) {
+      onChange(normalizedPool);
+    }
+  }, [onChange, pool]);
```
lucasbalieiro: feels like it will trigger an loop on rendering because  `onChange` is an inline callback from the parent, it will be a new reference on every render, causing this effect to fire on every render.

Just a hunch so far, I'm gonna do some manual tests to confirm

lucasbalieiro (commented):
I first of all thank you guys. Seems like a nice and structured improvement.
I've did the fist round of reading the important code changes here. I'm gonna take a time now to do some testing locally

lucasbalieiro (approved):
tACK looks very nice.

My concerned with loop useEffect was addressed when the `useRef` was addded in https://github.com/stratum-mining/sv2-ui/pull/218/changes/b44fe34690137ac67a822f6ab9bdb8bc505bb5a2

confirmed that the pool priority on the config file is correctly ordered after this refactor:

Image

confirmed that after a modifying the pools list inside the dashboard settings, the changes are applied correctly after tis refactor

Image

lucasbalieiro (approved):
Did a quick check on the main points after the rebase and LGTM

PR #225, opened 2026-07-21
Expose `min_individual_worker` hashrate config in settings

pavlenex:
Yesterday I was in a situation where I had to dynamically adjust lowest worker hashrate to debug an issue with hashrate rental. Since we don't expose this in settings, I had to re-do entire setup from scratch every time I wanted to do that which was quite annoying. 

This PR lets users update the lowest worker hashrate directly from Settings using the same logarithmic slider and unit-aware input found in the setup wizard. The slider is now a shared component to keep both screens visually and functionally consistent, while Settings uses a simplified version without miner presets. It also centralizes the default 100 TH/s value, validates user input, restarts services after saving, and ensures older configurations receive the correct default hashrate.

The 100 TH/s fallback was previously hardcoded in multiple places. This PR defines it once as DEFAULT_MIN_HASHRATE and reuses that constant in the setup UI and server configuration.

Screenshot 2026-07-21 at 13 16 24

### DIFF
--- a/server/src/config-generator.test.ts
+++ b/server/src/config-generator.test.ts
@@ -70,6 +70,7 @@ const NO_JD_DATA: SetupData = {
 test('translator config uses advanced setup values', () => {
   const config = generateTranslatorConfig(BASE_DATA_30);
 
+  assert.match(config, /min_individual_miner_hashrate = 100000000000000\.0/);
   assert.match(config, /downstream_extranonce2_size = 8/);
   assert.match(config, /shares_per_minute = 12\.5/);
 });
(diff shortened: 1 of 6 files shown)


lucasbalieiro (approved):
ACK 2d0b9895379e4867765def0a418f0d637038155b

GitGab19 (approved):
Pushed a small fix to the selector, which is now keeping the unit fixed until the focus is on: https://github.com/stratum-mining/sv2-ui/pull/225/commits/cf15f6f8b85528fada9117c9083aeeb114a9c9d8

PR #226, opened 2026-07-21
Refactor advanced mining settings into shared compact form

pavlenex:
This PR refactors the advanced mining configuration into a shared component used by both the setup wizard and Settings, ensuring the fields, defaults, validation, and descriptions remain consistent. It also introduces a more compact responsive layout by constraining the form width, placing Shares Per Minute and Extranonce2 Size side-by-side on wider screens, and keeping Coinbase Verification in a compact row while preserving mobile-friendly stacking.

My main motivation was `Shares Per Minute and Extranonce2 Size side-by-side` which inspired a bigger code unification. But tbh it's probably not worth the amount of changes, however if @lucasbalieiro likes it we can proceed with it, on my end.

Screenshot 2026-07-21 at 13 31 27

Screenshot 2026-07-21 at 13 30 21


lucasbalieiro (commented):
the field is allowing some "crazy" values that breaks the config file:

Image

translator dies cannot start due to bad config:
```bash
Translator proxy config error: Bad `config` TOML deserialize: `invalid type: 64-bit unsigned integer `1000000000000`, expected an unsigned 16 bit integer for key `downstream_extranonce2_size``
```

we probably already had this before, but worth fixing by adding a validation to the `Downstream extranonce2 ` field since we are already touching this area.

lucasbalieiro (approved):
ACK 7fefdfa2749727c0f9d23066028bbe519b358f8f

GitGab19:
Can you solve conflicts @pavlenex? I can do it if you prefer, but there's a chance I'd become the author of your commits after that.

lucasbalieiro (approved):
ACK rebase

PR #224, opened 2026-07-21
poc - JD port config

pavlenex:
This PR is proof of concept, it doesn't yet work for me so I hope at least UI or something can be useful. 

Screenshot 2026-07-21 at 12 54 22

```
2026-07-21T07:52:22.102358583Z [jdc] [stdout] 2026-07-21T07:52:22.102305Z  INFO jd_client_sv2: Job declarator client starting... setting up subsystems
2026-07-21T07:52:22.102553541Z [jdc] [stdout] 2026-07-21T07:52:22.102358Z  INFO jd_client_sv2: Initializing monitoring server on http://0.0.0.0:9091
2026-07-21T07:52:22.102557416Z [jdc] [stdout] 2026-07-21T07:52:22.102410Z  INFO jd_client_sv2: Using Bitcoin Core IPC socket at: /root/.bitcoin/node.sock
2026-07-21T07:52:22.102558416Z [jdc] [stdout] 2026-07-21T07:52:22.102420Z  INFO stratum_apps::monitoring::http_server: Starting monitoring server on http://0.0.0.0:9091
2026-07-21T07:52:22.102559166Z [jdc] [stdout] 2026-07-21T07:52:22.102424Z  INFO stratum_apps::monitoring::http_server: Cache refresh interval: 15s
2026-07-21T07:52:22.102559791Z [jdc] [stdout] 2026-07-21T07:52:22.102437Z  INFO jd_client_sv2::monitoring: Starting JDC miner telemetry loop with interval of 15 seconds
2026-07-21T07:52:22.102700875Z [jdc] [stdout] 2026-07-21T07:52:22.102637Z  INFO jd_client_sv2: Attempting to initialize upstream...
2026-07-21T07:52:22.102704666Z [jdc] [stdout] 2026-07-21T07:52:22.102643Z  INFO jd_client_sv2: Trying upstream 1 of 1: pool=sv2solo.ckpool.org:3336, jds=sv2solo.ckpool.org:3337
2026-07-21T07:52:22.103138041Z [jdc] [stdout] 2026-07-21T07:52:22.103037Z  INFO stratum_apps::monitoring::http_server: Swagger UI available at http://0.0.0.0:9091/swagger-ui
2026-07-21T07:52:22.103144666Z [jdc] [stdout] 2026-07-21T07:52:22.103055Z  INFO stratum_apps::monitoring::http_server: Prometheus metrics available at http://0.0.0.0:9091/metrics
[...]

PR #230, opened 2026-07-27
Allow custom pools (under JD mode) to specify a JDS port and update CK pool url

GitGab19:
This PR takes inspiration and replaces https://github.com/stratum-mining/sv2-ui/pull/224.

Successfully tested with the experimental JD endpoint of CKPool:
- url --> stratum.ckpool.org
- pool port --> 3336
- jd port --> 3337
- authority key --> 9anrRNhBh7869XtNnFcCuGBRZP51E635qGbu457J5kHdszhfRc3 

The only "problem" of CKPool is that it accepts only a max `extranonce_size` of 8bytes, and JDC requires 2bytes more for channel aggregation. So if you set the extranonce2 size to a value >6, JDC is gonna receive a `OpenExtendedMiningChannel.Error`, and it will fallback.

### DIFF
--- a/server/src/config-generator.test.ts
+++ b/server/src/config-generator.test.ts
@@ -154,6 +154,7 @@ test('jdc config uses shared shares-per-minute and miner signature', () => {
 
   assert.ok(config);
   assert.match(config, /shares_per_minute = 12\.5/);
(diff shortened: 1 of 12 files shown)


pavlenex (approved):
tACK on both JD and non JD mode of CK Pool
Screenshot 2026-07-28 at 18 21 18

PR #232, opened 2026-07-28
feat(ui): streamline dashboard and surface per-worker best diff

GitGab19:
This PR introduces improvements about best diff, and cleans up some unnecessary text and columns from the dashboard.

https://github.com/user-attachments/assets/7fdc7540-2d55-4aeb-84da-4997df4c1bde


pavlenex (approved):
cACK

PR #234, opened 2026-07-29
make generated service configs update-safe

GitGab19:
This PR regenerates managed Translator/JDC configs from saved setup before starts and auto-starts, so app updates cannot keep stale TOML files around.

If saved setup is valid but missing required choices for the current version, surface a setup review with existing values prefilled. If saved setup cannot be read, preserve it and require an explicit reset. Also add shared validation and a release smoke check for generated configs against the selected `sv2-apps` images.

It unblocks https://github.com/getumbrel/umbrel-apps/pull/5933.

Here a recording of the exact case described in the PR reviewed by the Umbrel team, where a TOML config file is stale, and it gets recovered by using the `state.json` information:

https://github.com/user-attachments/assets/aeee4ac2-e021-437b-8e52-e2a8581ae209

Here a recording of the case where the user runs it from an incomplete state:

https://github.com/user-attachments/assets/b7eb219d-cd6f-4259-8ca5-5ad75384db62

Here a recording of the case where the user runs it from an invalid `state.json`:

https://github.com/user-attachments/assets/2518d598-7216-4615-93b9-3e94469784f3


lucasbalieiro (commented):
concept ACK  82c3dd99744bcc2b6942c63fb6029ae8e12f13bd

server/src/index.ts
```diff
@@ -147,45 +94,23 @@ function configuredPools(data: SetupData): PoolConfig[] {
...
-  }
+/**
+ * Stop the stack, reconcile generated service configuration from saved setup
+ * data, persist that data, then start the stack. Every user-initiated start
+ * path goes through this function so updates cannot retain stale TOML files.
+ */
+async function reconcileAndStartStack(prepared: Extract<PreparedServiceConfig, { kind: 'ready' }>): Promise<void> {
+  // Preparation happens before disrupting an active stack. If saved input
+  // needs review, callers return that guidance without causing downtime.
+  await stopStack();
```
lucasbalieiro, line 110: maybe would be wiser to do the `stopStack` after the config reconciliation.

Because if something goes wrong during the TOML generation, we will render the user with a stopped stack and a corrupted toml file.

lucasbalieiro: oh, I'm wrong, the callers of this are  using`prepareServiceConfig`. 

so the comment above this is not lying.

lucasbalieiro (approved):
ACK 82c3dd99744bcc2b6942c63fb6029ae8e12f13bd

I've tried to test this locally by running an older version of `sv2-ui` (0.5.0 and 0.4.0), configured the wizard to a sovereing solo mining mode and then running again with the changes of this PR (simulating an update of the app) and the transition went well,

PR #239, opened 2026-08-06
fix: remove shell injection from socket existence check

lucasbalieiro:
closes https://github.com/project-loupe/audit-sv2-ui/issues/6

needs testing on linux to see if the node.sock probe still working from docker and dev envs

### DIFF
--- a/server/src/bitcoin-socket-exists.test.ts
+++ b/server/src/bitcoin-socket-exists.test.ts
@@ -1,0 +1,52 @@
+import assert from 'node:assert/strict';
+import { test } from 'node:test';
+import { execFileSync } from 'node:child_process';
+import { existsSync, mkdtempSync, rmSync, writeFileSync } from 'node:fs';
+import { tmpdir } from 'node:os';
+import path from 'node:path';
+import { bitcoinSocketExistsScript } from './bitcoin-socket-exists.js';
+
+test('bitcoinSocketExistsScript must not execute shell commands embedded in the socket path', () => {
+  const tmp = mkdtempSync(path.join(tmpdir(), 'sock-inj-'));
+  const marker = path.join(tmp, 'pwned');
+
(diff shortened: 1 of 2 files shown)


server/src/bitcoin-socket-exists.test.ts
```diff
@@ -0,0 +1,47 @@
...
+      stdio: 'ignore',
+    });
+    assert.fail('expected non-zero exit for missing socket');
+  } catch (err) {
+    assert.ok(
+      err && typeof err === 'object' && 'status' in err && err.status !== 0,
+      'missing socket should exit non-zero',
+    );
+  }
+});
```
GitGab19, line 52: Clanker suggestion:

```
This test name says it covers both success and failure, but it only asserts the missing-path failure case. Could we also create the path and assert the script exits successfully? That would catch regressions where the probe always returns non-zero.
```

```suggestion
test('bitcoinSocketExistsScript exits 0 when the socket path exists and non-zero otherwise', () => {
  const tmp = mkdtempSync(path.join(tmpdir(), 'sock-exists-'));
  try {
    const socketPath = path.join(tmp, 'node.sock');

    assert.throws(
      () => execFileSync(process.execPath, ['-e', bitcoinSocketExistsScript, socketPath], {
        stdio: 'ignore',
      }),
      /Command failed/,
      'missing socket path should exit non-zero',
    );

    writeFileSync(socketPath, '');

    execFileSync(process.execPath, ['-e', bitcoinSocketExistsScript, socketPath], {
      stdio: 'ignore',
    });
  } finally {
    rmSync(tmp, { recursive: true, force: true });
  }
});
```

lucasbalieiro: amended the suggestion

GitGab19 (approved):

PR #243, opened 2026-08-13
force manual setting of username when in Pool Mode

lucasbalieiro:
closes https://github.com/project-loupe/audit-sv2-ui/issues/15 
took the chance to fix other bugs I've found during the process


GitGab19:
> out of context: I was able to reproduce this on Solo Mode. but this is an expected behavior. Since on Solo Mode the username is inherited.

Here's a recording of a POOL mining setup, started from scratch, where I initially select only Braiins Pool, configure its username, and then I go back, I add a Custom Pool as a fallback, and I can click the "Continue" button without having to configure the username of the fallback pool:


https://github.com/user-attachments/assets/d5f9a2a3-3255-424f-8025-c25a1fabcd13


image

lucasbalieiro:
@GitGab19, sorry. I was unable to reproduce the things you showed in your videos.

I've  just rebased and solved the conflicts

GitGab19 (approved):
I just tested this again, and the issues I was mentioning are gone

Good job @lucasbalieiro !

PR #244, opened 2026-08-17
Add input validation

### LINKED ISSUE #242: pool address input needs validation
Image

Ayush4958:
### Description
I had added input validation for custom pool addresses to prevent common user configuration errors 

**Closes #242**

### Changes
- Added validation logic to `src/lib/utils.ts` to strictly validate standard hostnames, IPv4, and IPv6 addresses.
- **State Normalization**: The address field now automatically `.trim()`s leading trailing whitespace on change before saving to state.
- Added unit tests covering valid inputs, protocols, spacing edge cases & invalid characters.


lucasbalieiro (dismissed):
*APPROVED BY MISTAKE., I've selected the wrong button. sorry 😢 *

can you make the error message appear only when the user start typing or move the focus out of the input without filling it?

Image

Ayush4958:
> *APPROVED BY MISTAKE., I've selected the wrong button. sorry 😢 *

> can you make the error message appear only when the user start typing or move the focus out of the input without filling it?

Ohk sure, I will fix this one

src/lib/utils.ts
```diff
@@ -122,3 +122,27 @@ export function getPoolAuthorityPubkeyError(v: string): string | null {
...
+export function isValidPoolAddress(v: string): boolean {
+  if (!v) return false;
+  
+  const normalized = v.trim();
+  if (normalized.includes('://')) return false;
+  if (normalized.includes('@')) return false;
+  if (normalized.includes(',')) return false;
+  if (/\s/.test(normalized)) return false;
+
+  const VALID_CHARS_REGEX = /^[a-zA-Z0-9\-.:_]+$/;
```
lucasbalieiro: I would not only validate the charset.

Because this validation of the charset only will accept inputs like: :::::::::, ...., pool.com:3333 (this will lead to an invalid config file), 1.2.3.4.5, 192.168.001.1, pool..com, -pool.com.


We should do some sort of "struct-aware" validation. And move the validation to `@shared`

lucasbalieiro: MAYBE, there is some lib that already do that.

Ayush4958: Gotcha, will look in lib 
If i can find something that works for this

GitGab19:
Clanker review:

---

The `validator` switch does what @lucasbalieiro asked for: every case from the first review round (`:::::::::`, `....`, `pool.com:3333`, `1.2.3.4.5`, `192.168.001.1`, `pool..com`, `-pool.com`) is now rejected, hostnames / single-label names (`localhost`) / IPv4 / unbracketed IPv6 are accepted, and the error only appears after typing or blur as requested. Nice work.

There are three things I'd like addressed before merging, plus a few smaller notes.

## Blocking

### 1. The error is only cosmetic — an invalid address still reaches the generated config

Both the wizard's **Continue** (`PoolConfigStep`, `disabled={!isValid}`) and the settings **Save** (`ConfigurationTab`, `saveDisabled={!arePoolsValid}`) bottom out in `isPoolConnectionComplete` in `src/lib/poolValidation.ts`, which only checks `pool?.address` for truthiness. Server-side, `getPoolConfigError` in `server/src/pool-validation.ts` only checks TOML-safety. So `pool.com:3333` turns the field red, but the user can still click through, and the server writes `address = "pool.com:3333"` — exactly the broken config Lucas warned about.

Since you already moved the function to `@sv2-ui/shared`, this is a two-line fix:

- `src/lib/poolValidation.ts` → replace `pool?.address &&` with `isValidPoolAddress(pool?.address ?? '') &&`
- `server/src/pool-validation.ts` → extend the address check with `|| !isValidPoolAddress(pool.address)` (and add a `pool.com:3333` case to `pool-validation.test.ts`)

### 2. `import validator from 'validator'` adds ~210 kB to the bundle

The default import pulls in validator's CJS `index.js`, which drags the entire library into the bundle (`isCreditCard`, `isIBAN`, `isMobilePhone`, … are all in `dist/`). Measured: `index-*.js` goes from ~1,479 kB to ~1,688 kB (gzip 322 kB → 370 kB) for two functions.

Switching to deep imports fixes it completely:

```ts
import isIP from 'validator/lib/isIP';
import isFQDN from 'validator/lib/isFQDN';
```

I tested this locally: the bundle goes back to ~1,476 kB with zero validator leftovers, typecheck passes, all tests pass, and it works under `tsx` on the server (`@types/validator/lib/*.d.ts` exist for both).

### 3. Needs a rebase onto `main`

The branch is 17 commits behind and conflicts in `shared/package.json`, `shared/src/validation.ts`, `src/lib/utils.ts` and `src/lib/utils.test.ts` — #245 moved `stripWrappingQuotes` / `isValidPoolAuthorityPubkey` into shared, so both branches append to the same files. Straightforward to resolve, but please rebase (not merge) and force-push.

## Non-blocking

- **Add regression tests for the cases that motivated the library switch.** `pool.com:3333`, `pool..com`, `-pool.com`, `192.168.001.1`, `1.2.3.4.5` all pass today but aren't in `utils.test.ts`. Worth locking in.
- **Non-ASCII slips through.** `isFQDN` accepts `pöol.example.com` and `pool.com\u200b` (a zero-width space survives both `.trim()` and `/\s/`), and SRI won't IDNA-encode them. A guard like `if (!/^[\x21-\x7e]+$/.test(normalized)) return false;` closes that and also subsumes the whitespace check.
- **Duplication.** `getPoolAddressError` re-implements every check in `isValidPoolAddress` and then calls it anyway. Simpler to make `isValidPoolAddress` just `getPoolAddressError(v) === null`. Related: a whitespace-only string returns "Invalid address format" instead of "Address is required" — unreachable from the UI thanks to the trim, but the shared function will be used server-side too.
- **Touched-state resets on reorder.** `isAddressTouched` lives in `CustomPoolFields`, whose row key is `selected-pool-${index}`, so dragging the custom pool remounts it. Only affects the empty-address "required" message, so fine to leave.
- **Noise.** There are unrelated whitespace-only edits in `src/lib/utils.ts`, and two commits share the same message — please drop the former and squash the latter when you rebase. (The `"peer": true` removals in `package-lock.json` are fine — current npm strips those regardless of who regenerates the lockfile.)

GitGab19 (approved):
@lucasbalieiro can you take a final look at this one?

lucasbalieiro (commented):
ACK, needs rebase

Ayush4958:
> ACK, needs rebase

Rebased the branch & Squashed the commits

lucasbalieiro (approved):

PR #245, opened 2026-08-17
address multiple loupe issues

lucasbalieiro:
closes https://github.com/project-loupe/audit-sv2-ui/issues/8
closes https://github.com/project-loupe/audit-sv2-ui/issues/13
closes https://github.com/project-loupe/audit-sv2-ui/issues/17
closes https://github.com/project-loupe/audit-sv2-ui/issues/20
closes https://github.com/project-loupe/audit-sv2-ui/issues/3
closes https://github.com/project-loupe/audit-sv2-ui/issues/5
closes https://github.com/project-loupe/audit-sv2-ui/issues/7
closes https://github.com/project-loupe/audit-sv2-ui/issues/21
closes https://github.com/project-loupe/audit-sv2-ui/issues/35
closes https://github.com/project-loupe/audit-sv2-ui/issues/34
closes https://github.com/project-loupe/audit-sv2-ui/issues/33

### DIFF
--- a/server/src/active-pool.test.ts
+++ b/server/src/active-pool.test.ts
@@ -168,3 +168,46 @@ test('tracker carries a pending connection across incremental polls', async () =
     index: 1,
   });
 });
+
+test('invalidates the active pool when the authority key rotates without a new handshake', async () => {
+  let readCount = 0;
+  const tracker = new ActivePoolTracker(async (_container, options) => {
+    readCount += 1;
+    if (readCount === 1) {
+      // Initial full read under the original key: a successful handshake.
+      return [
+        log('Trying upstream 2 of 2: fallback.example.com:4444'),
+        log('Connected to upstream at 192.0.2.42:4444'),
+        log('Received: SetupConnectionSuccess(used_version: 2, flags: 0x00000000)'),
+      ];
+    }
+    // Second interaction.
+    if (options === undefined) {
+      // Fix path: the key change forces a full re-read, which now also shows
+      // the new (not-yet-successful) attempt after the rotation.
+      return [
+        log('Trying upstream 2 of 2: fallback.example.com:4444'),
+        log('Connected to upstream at 192.0.2.42:4444'),
+        log('Received: SetupConnectionSuccess(used_version: 2, flags: 0x00000000)'),
+        log('Trying upstream 2 of 2: fallback.example.com:4444'),
+      ];
+    }
+    // Buggy path: unchanged configKey means an incremental read; no new logs
+    // have appeared yet, so the stale active index would otherwise survive.
+    return [];
+  });
+
+  const poolsWithKey = (key: string): PoolConfig[] => [
+    { ...POOLS[0], authority_public_key: key },
+    { ...POOLS[1] },
+  ];
+
+  // First: connected under the original key.
+  assert.deepEqual(await tracker.getActivePool('translator', poolsWithKey('primary-key')), {
+    name: 'Fallback',
+    index: 1,
+  });
+
+  // Then the authority key rotates (same endpoint) without a new handshake.
+  assert.equal(await tracker.getActivePool('translator', poolsWithKey('rotated-key')), null);
+});
--- a/server/src/active-pool.ts
+++ b/server/src/active-pool.ts
@@ -111,7 +111,13 @@ export function detectActivePool(
 function getConfigKey(container: LogContainerRole, pools: PoolConfig[]): string {
   return JSON.stringify([
     container,
-    ...pools.map((pool) => [pool.name, normalizeHost(pool.address), pool.port]),
+    ...pools.map((pool) => [
+      pool.name,
+      normalizeHost(pool.address),
+      pool.port,
+      pool.jds_port,
+      pool.authority_public_key,
+    ]),
   ]);
 }
 
--- a/server/src/atomic-write.test.ts
+++ b/server/src/atomic-write.test.ts
@@ -1,0 +1,51 @@
+import assert from 'node:assert/strict';
+import { chmod, mkdtemp, rm, stat, writeFile } from 'node:fs/promises';
+import os from 'node:os';
+import path from 'node:path';
+import { test } from 'node:test';
+import { writeFileAtomically } from './atomic-write.js';
+
+async function withRestrictiveUmask(run: () => Promise<void>): Promise<void> {
+  const previousUmask = process.umask(0o077);
+  try {
+    await run();
+  } finally {
+    process.umask(previousUmask);
+  }
+}
+
+test('preserves an existing 0644 mode under a restrictive umask', async () => {
+  const dir = await mkdtemp(path.join(os.tmpdir(), 'sv2-ui-atomic-'));
+  const file = path.join(dir, 'config.toml');
+  try {
+    await writeFile(file, 'original');
+    await chmod(file, 0o644);
+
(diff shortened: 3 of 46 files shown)


GitGab19:
Clanker review:

## PR review: issues to resolve

### Required before merge

- [ ] **Apply authority-key length validation on the server**

  The PR validates that a decoded authority key is 34 bytes in `src/lib/utils.ts`, but the independent validator in `server/src/pool-validation.ts` still accepts any checksum-valid Base58Check payload.

  A direct API request can therefore bypass the frontend validation and submit a malformed authority key.

  **Suggested fix:**

  - Share the authority-key validator between the frontend and server, or add `decoded.length === 34` to the server validator.
  - Add a server regression test using checksum-valid payloads of incorrect lengths.

- [ ] **Bind official pool branding to the authority public key**

  `getKnownPoolForConfig()` currently identifies known pools using only `address` and `port`. A configuration using an official endpoint but an attacker-controlled `authority_public_key` is still shown with official branding.

  **Suggested fix:**

  - Include the active pool’s authority public key in `/api/status`, or return a server-derived trusted preset ID.
  - Match `address`, `port`, and `authority_public_key`.
  - Include `jds_port` where it is relevant to pool identity.
  - Add a regression test proving that the correct endpoint with the wrong authority key receives no official branding.

### Recommended improvements

- [ ] **Canonicalize miner telemetry CIDRs**

  The stricter parser still accepts noncanonical forms such as:

  ```text
  192.168.001.000/024

lucasbalieiro:
@GitGab19, addressed the points

GitGab19:
## Follow-up review — issues to resolve

The current test suite, lint, and builds pass, but the following behavioral issues remain.

### Required before merge

- [ ] **Reject quoted authority keys at the server boundary**

  `isValidPoolAuthorityPubkey()` strips wrapping single quotes before decoding, while `isTomlSafeIdentifier()` allows single quotes. As a result, the server accepts values such as:

  ```text
  '9awtMD5KQgvRUh2yFbjVeT7b6hjipWcAsQHd6wEhgtDT9soosna'
  ```

  The generated TOML retains those quotes, so Translator/JDC receive an invalid authority key at startup.

  **Suggested fix:**

  - Require the server-side value to already be canonical, for example `stripWrappingQuotes(value) === value`, before validating it.
  - Add regression tests covering both `PUT /api/config` and `POST /api/setup`.

- [ ] **Fix the SRI worker-name `/` validation flow**

  The early return in `updateSriIdentity()` leaves the invalid worker name only in component-local state. This causes several problems:

  - Payout-address and donation changes are discarded while the worker contains `/`.
  - Continue/Save remains enabled because validation sees the previous valid `pool.user_identity`.
  - The visible form and the value eventually saved can differ.
  - Existing identities such as `.../rig/1` are silently normalized to `.../rig1` when the editor mounts.

  **Suggested fix:**

  - Integrate worker-name validation into the form’s existing validity calculation, preferably through `getSriIdentityError()`.
  - Remove the early return/local-state divergence.
  - Avoid silently rewriting existing worker names.
  - Add component-level tests covering Continue/Save, address changes, and donation changes while the worker name is invalid.

- [ ] **Preserve the exact existing file mode during atomic writes**

  The mode passed to `fs.open()` is still filtered by the process umask. For example, an existing `0644` file becomes `0600` under umask `077`.

  **Suggested fix:**

  - Call `await handle.chmod(mode)` after opening the temporary file when an existing mode was found.
  - Add regression tests under restrictive umasks.

- [ ] **Apply persistence bounds to every dashboard metric**

  The new limits cover only `share_stats`. `blocks_found` and `best_diff` still:

  - accept an unlimited number of telemetry-derived keys;
  - load storage without validation or a size limit;
  - continually clone and serialize the growing object;
  - use `Math.max(...Object.values(...))`, which eventually throws a `RangeError` and can make the dashboard crash on every reload until storage is cleared.

  **Suggested fix:**

  - Apply entry-count, key-length, value, and storage-size normalization to `usePersistentMetric()`.
  - Replace the `Math.max(...values)` spread with a bounded `reduce()`.
  - Hash or otherwise bound channel identity keys instead of silently dropping identities that exceed 256 characters.
  - Preserve evicted share counts in an aggregate bucket so lifetime totals cannot decrease.
  - Add regression tests for oversized stored state, channel churn, and monotonic totals.

- [ ] **Invalidate active-pool state when the authority key changes**

  `ActivePoolTracker.getConfigKey()` currently includes the pool name, address, and port, but not `authority_public_key`. After a configuration restart, the tracker can therefore retain an old successful `activeIndex` and associate it with the new key before any new Noise handshake succeeds.

  This can make `/api/status` report the pool as connected and display trusted branding based on stale connection state.

  **Suggested fix:**

  - Include `authority_public_key` and relevant connection fields such as `jds_port` in the tracker configuration key, or explicitly reset the tracker whenever configuration changes.
  - Add a regression test proving that changing the key without a new successful handshake results in no active pool.

- [ ] **Make HSL validation and parsing agree**

  `HSL_TRIPLET_RE` accepts arbitrary whitespace through `\s+`, while the consumers parse with `split(' ')`.

  Examples:

  - Double spaces produce incorrect parsed components.
  - Tab-separated values pass validation but can crash the Settings page in `hslToHex()`.

  **Suggested fix:**

  - Create one parser that validates and returns the three numeric components.
  - Reuse it in `loadConfig()`, `applyCssVariables()`, and `hslToHex()`.
  - Validate the actual ranges: hue `0–360`, saturation/lightness `0–100`.
  - Add tests for tabs, repeated spaces, malformed values, and out-of-range values.

### Recommended completeness improvements

- [ ] **Separate trusted-pool identity from endpoint matching**

  Strict address/port/authority-key matching is correct for trusted branding, but the same `isSamePool()` function is also used for editor selection and deduplication.

  Consider separate helpers for:

  - strict trusted identity matching;
  - endpoint-only duplicate detection;
  - deciding whether duplicate endpoints with different keys are intentionally supported.

  Endpoint-only matching should not be used to display official branding for a wrong-key configuration.

- [ ] **Stop using the user-controlled pool name to select dashboard behavior**

  `isAggregatedTproxyPoolName(configPoolName)` still changes the dashboard layout based on the configured display name.

  Derive this behavior from the authenticated pool tuple or a server-derived capability/preset identifier. The existing `shouldAggregateTranslatorChannels()` helper is not strict enough as written because it matches the address **or** authority key.

- [ ] **Validate custom logos before storing them**

  The file input’s `accept="image/*"` is advisory. A non-image file can currently be stored and previewed as broken, while the header rejects it and a reload later discards it.

  Validate the file/data URL at upload time, surface an error, and only persist accepted image values.

lucasbalieiro:
@GitGab19, I'm not sure if I understood what the AI is saying. The thing is getting too complex 😅 .

I've push two separated commits to address this last comment you've made because I'm not sure if going in the same direction that the AI is pointing, so I can easily drop them if needed ([ad5ea24](https://github.com/stratum-mining/sv2-ui/pull/245/commits/ad5ea24698cdc5dafcd7190ce3f4d625f481bcb3) and [b20503b](https://github.com/stratum-mining/sv2-ui/pull/245/commits/b20503bfdf703334e33efc400bfd34368c367c69)).

GitGab19 (approved):
@lucasbalieiro we can proceed with this PR as it is now.

I opened https://github.com/stratum-mining/sv2-ui/issues/246 for a follow-up PR.

PR #247, opened 2026-08-27
fix(docker): refine offline socket error message

### LINKED ISSUE #179: improve error message when dock.sock is unavailable
during the setup process (IPC validation), if the communication with `docker.sock` fails we get an ugly error:

Image

Ayush4958:
### What this PR does
Improves the error message shown in the Setup Wizard when the Docker daemon is unreachable or the `DOCKER_SOCKET_PATH` is misconfigured.

**Closes #179** 

### Changes
- Removed the raw, verbose `ECONNREFUSED` debug string from surfacing to the UI.
- The error now concisely instructs the user to start Docker or check their path, and cleanly lists available sockets only if they exist on the system.


Ayush4958:
Hi @GitGab19 ,
PR is ready for review 
take time to review it 

thanks ....

GitGab19:
If you use the wrong syntax in the description, the issue you reference won't get closed, please fix it.

Also, why 4 commits? The first two seems to be doing the same thing? Maybe they should be squashed.

Ayush4958:
> If you use the wrong syntax in the description, the issue you reference won't get closed, please fix it.
> 
> Also, why 4 commits? The first two seems to be doing the same thing? Maybe they should be squashed.

ohk,  I will do it

Ayush4958:
@GitGab19 ,
Updated both things u mentioned

GitGab19:
Clanker review:

---

Thanks, the change of routing the probe errors through `normalizeDockerError` is good and it's what actually fixes #179.

The new message though has a problem: the sockets list is not filtered against the endpoint we just failed on. Since the socket is auto-detected by default, in the most common case (Docker stopped, socket file still there) the user gets this:

`Docker is not reachable at /var/run/docker.sock. Available sockets found: /var/run/docker.sock. Try setting DOCKER_SOCKET_PATH to one of these.`

so we suggest to set `DOCKER_SOCKET_PATH` to the same socket which just failed, and the "make sure Docker is running" hint (which is the actual fix) is gone in that branch. Same message for `EACCES` (user not in the docker group), where changing the path can't help.

Also `DOCKER_HOST` disappeared from the message, but README documents it as an override together with `DOCKER_SOCKET_PATH`, and I would keep the `(${source})` part, it's short and it tells the user which setting picked that endpoint.

Could you:
- filter out `dockerConnection.endpoint` from the list, and only suggest `DOCKER_SOCKET_PATH` when there are *other* sockets
- always keep the "make sure Docker is running" hint, and mention `DOCKER_HOST` too
- (optional) a specific hint for `EACCES`/`EPERM`
- adjust the "available sockets" test accordingly: right now it mocks `DOCKER_SOCKET_PATHS[0]` which is also the default endpoint, so it's exactly the case which should *not* be listed. Mock a different path and assert the endpoint is not in the list.

Ayush4958:
Added all changes requested 
It's ready for review 
take time to review it 

thanks ...

GitGab19 (commented):
Second round of clanker review:

---

Thanks, this looks good now, the messages are correct in all the cases I tried.

Two small nits:

- the "filters current endpoint" test doesn't actually test the filter: the mock already excludes `/var/run/docker.sock`, so the filter is a no-op there (I removed it and the test still passes). Mock `existsSync` to always return `true` and assert the endpoint is not in the part after `Other available sockets found:`, that one fails without the filter.
- `Try setting DOCKER_SOCKET_PATH or DOCKER_HOST to one of these`: `DOCKER_HOST` needs the `unix://` prefix, with a bare path the server fails at startup with `Invalid URL`. So in that branch I would only suggest `DOCKER_SOCKET_PATH`, and keep `DOCKER_HOST` in the generic one (where it already is).

Also please remove the trailing whitespace on the blank lines.

Ayush4958:
Fixes the mention changes 
PR is ready for another round of review 

thanks ...

GitGab19 (approved):
ACK, can you squash commits into a single one?

Ayush4958:
Squashed all commits into a single one ....

PR #248, opened 2026-08-29
add authenticated principal and remove wildcard CORS

lucasbalieiro:
closes https://github.com/project-loupe/audit-sv2-ui/issues/26
closes https://github.com/project-loupe/audit-sv2-ui/issues/1

I'm introducing authentication for the API routes that gate `/api/*`, including both control routes and operational frontend routes.

removed the unauthenticated `cors()` middleware and replaces it with dynamic same-origin enforcement that works across localhost, Docker, and Umbrel without relying on a static allowlist.

the API should no longer be accessible from arbitrary websites that the user happens to visit. The `/`, `/setup`, and `/settings` routes now require an authenticated session, while `/faq` remains public.

### POTENTIAL BREAKING CHANGE !!!!!

Let's consider the upgrade flow for an existing user.

If a user upgrades to this version, they will **NOT** need to redo their existing configuration for the sv2-apps. Instead, they will be prompted to create a password through a screen like this:

image

After creating the password, a **Recovery Key** will be displayed. The user can use this key to reset their password if they forget it:

image

Resetting the password does **not** delete or modify the existing `sv2-apps` configuration.

If the user loses both their password and recovery key, there is no way for us to recover access. They will need to uninstall and reinstall the application and go through the setup process again.

I've added an `uninstall` command to the README so Docker users have a straightforward way to completely remove the application and start over.

During the review, I think we should avoid focusing exclusively on the "clanker review" aspect of this PR. This introduces a new authentication and user-facing flow, so reviewing only the implementation details may not be sufficient. The overall behavior and upgrade/recovery experience should also be considered.

At the same time, I think we should keep the threat model and the complexity of the solution proportional to the problem we're solving.

We are protecting a locally hosted application and its API, not designing an authentication system for a globally exposed financial service. This is why I intentionally avoided introducing more complex recovery mechanisms such as SMS/email recovery, elaborate cryptographic schemes, asymmetric encryption, or other infrastructure that would significantly increase the complexity of the project.

The goal here is to prevent unauthorized access to the local API and protect the application from the attack scenarios identified during the security review. I believe this approach provides a reasonable security boundary without unnecessarily increasing the complexity of `sv2-ui`.

There are still a few things I need to test to make sure I haven't introduced any problems. I'm currently traveling, so I don't have access to my Bitaxes or Umbrel machine to test these scenarios yet:

* [ ] do manual testing for Bitaxe telemetry.
* [ ] add an "uninstall" command in the FAQ and make the FAQ public when inserting or creating the password
* [ ] probably gonna need some slightly different workflow on Umbrel, but unsure
* [ ] Test with Umbrel to verify that the proxy integration continues to work correctly.


src/components/auth/LoginGate.tsx
```diff
@@ -0,0 +1,331 @@
...
+  onDone,
+}: {
+  recoveryKey: string;
+  onDone: () => void;
+}) {
+  const [copied, setCopied] = useState(false);
+
+  const copy = async () => {
+    try {
+      await navigator.clipboard.writeText(recoveryKey);
```
lucasbalieiro: this is not Working on Umbrel.

Probably because of non-HTTPs

lucasbalieiro: the copy paste of the stratumv2 url is working on the dashboard. 

I should draw some inspiration from there

lucasbalieiro: fixed

got the copy mechanism from the dashboard and transformed into a component and re-used in the two new places I'm introducing in this PR: check file `src/components/ui/copyable-value.tsx`

server/src/index.ts
```diff
@@ -68,10 +90,83 @@ function isJsonObject(value: unknown): value is Record<string, unknown> {
   return typeof value === 'object' && value !== null && !Array.isArray(value);
 }
 
-// Middleware
-app.use(cors());
```
lucasbalieiro: I've left my umbrel running all night with a bitaxe plugged to it.

And this new adaptations to the `cors` flow that I've introduced did not cause any trouble:
- updating existing configs worked;
- I was able to fetch the bitaxe telemetry; 
- the new configuration `credentials.json` did not impact the existing configuration files on the umbrel app;
- the checks on `secFetchSite`, `Origin` and others did not caused a regression on Umbrel deployment where the client and the server are not in the same machine. So `cors` is not a problem anymore

lucasbalieiro:
Added two new commits for the scenario where a user loses their recovery key:

* [Make the FAQ visible from the auth gates](https://github.com/stratum-mining/sv2-ui/pull/248/commits/f40d46af0a9ae994b9572550b4633defa787c77c)
* [Add scrolling to FAQ answers and a new question about the recovery key](https://github.com/stratum-mining/sv2-ui/pull/248/commits/04547a342c3d70bc1279c08f4562007cb64347f7)

If a user loses their recovery key, there is still a last-resort recovery path, but it requires filesystem access. This is intentional: I want to ensure that the person performing the reset has the appropriate privileges, rather than providing a frontend-only alternative that could bypass the existing credential settings.

The goal is to make recovery possible without weakening the security model. If a user loses their recovery key and also does not have access to the filesystem, there is no recovery path available through the application.

The only remaining option would be to uninstall and reinstall the application, which would reset the configuration. However, that also requires the necessary privileges to manage the application.

If someone has neither the recovery key nor the privileges required to reset or reinstall the application, I think we should not provide another way to bypass the password.

GitGab19:
Hey @lucasbalieiro, I've been playing around with this PR, and I think what you did is great.

I have some comments/suggestion about the UX:

- nothing in the client reacts to a 401 mid-session, so after 12h the dashboard just freezes until you switch tabs
- every sv2-ui restart logs everyone out (in-memory sessions). Fine for a laptop, but on Umbrel this happens on every update/reboot and the dashboard is often left open on a screen. I'm ok with it if we're explicit about it in the release notes, but maybe a sliding session would help too
- the onboarding copy still says losing the recovery key means reinstalling / wiping the volume, which contradicts the new FAQ entry
- upgrading users see "Create an admin password" with nothing telling them their setup is intact
- no way to change the password while logged in
- I'd add the usual show/hide toggle on the password fields

On the README: I don't think we should present the uninstall command as the answer to a lost recovery key. The FAQ path keeps the mining config and needs the same host access, so I'd point the README at that instead, and keep uninstall only as a general cleanup command if we want one.

GitGab19:
Besides of the UX points, Fable found some issues on the code, which I'm gonna report here:

### Blocker: auth can be bypassed by changing the case of the path

Express matches routes case-insensitively, but the guard in `index.ts` checks `req.path.startsWith('/api/')` case-sensitively. So with a password set and no cookie:

```
GET /api/env                  -> 401
GET /API/env                  -> 200
GET /Api/config               -> 200
GET /TRANSLATOR-API/v1/global -> 502 (it reaches the proxy)
```

Every protected route is reachable like this, including setup/reset/stop/restart, so anyone on the LAN with curl has the full API. I think the fix is to mount the auth middleware with the same matcher express uses for routes, something like `app.use(['/api', '/translator-api', '/jdc-api'], requireSession)` with the public paths allowlisted inside, or enable case-sensitive routing. Either way we need a regression test that actually spawns the server (like `auth-recovery.test.ts` does), the regex checks over the source in `index.test.ts` can't catch this.

### Rate limiting

- `app.set('trust proxy', true)` makes `req.ip` whatever the client puts in `X-Forwarded-For`. I sent 7 wrong recovery keys with different forged IPs and never got a 429, while from the same IP it kicks in after 5. The failures map also grows without bound this way.
- Login has no rate limit at all, 8 wrong passwords in a row all came back 401 with no throttling.

Since this is a single-operator app I'd just use a global limiter that ignores the IP, for both login and recover. If you prefer per-IP, trust proxy should be limited to `'loopback, linklocal, uniquelocal'`.

### Recovery leaves the instance unowned

`/api/auth/recover` deletes the credential and returns, and the new password is set in a separate request. In between, `passwordSet` is false and whoever hits `setup-password` first owns the instance. I'd accept `{ recoveryKey, newPassword }` in the recover request and write the new credential (with a fresh recovery key) in one go. The same window exists on upgrade for every already configured instance, so I think we should call that out clearly in the release notes.

### same-site is treated as same-origin

`isSameOriginRequest` accepts `Sec-Fetch-Site: same-site` and the Origin fallback ignores the port. `SameSite=Strict` cookies are also sent on same-site requests. On Umbrel all apps share the host and differ only by port, so another app open in the same browser could drive sv2-ui with the user's session. I'd accept only `same-origin` and `none`, and compare the full origin including port in the fallback. The vite proxy already looks same-origin to the browser so dev keeps working.

### The client falls open on auth-state errors

`LoginGate` renders the app on any error from `/api/auth/state`, including a 500 from a malformed `credential.json` or a transient error while the server restarts. Not a real bypass since the API still 401s, but the user ends up on a broken dashboard with no way back to the login screen. I'd treat only network failure / 404 as standalone.

### Minor

- `parseCookies` throws on malformed input, I got a 500 with `Cookie: sv2_ui_session=%E0%A4%A`. The token is base64url so I think we can skip `decodeURIComponent` there.
- `GET /api/auth/recovery-key` duplicates `recoveryKeySet` from `/api/auth/state` and nothing in the client calls it.
- `saveCredential` writes the file with default perms and chmods to 0600 afterwards, passing the mode to the atomic writer would avoid the window.

Everything else looks good to me: scrypt with per-credential salt and constant-time compare, HttpOnly cookie, recovery key shown once, reset wiping credential + sessions, and the translator/jdc proxies finally behind auth. Tests, typecheck and lint are all green on my side.

lucasbalieiro:
rebased

about comments: https://github.com/stratum-mining/sv2-ui/pull/248#issuecomment-5617500441 and https://github.com/stratum-mining/sv2-ui/pull/248#issuecomment-5617566898

> nothing in the client reacts to a 401 mid-session, so after 12h the dashboard just freezes until you switch tabs

that is a good point, addressed in a new commit [d6a282d](https://github.com/stratum-mining/sv2-ui/pull/248/commits/d6a282df7f71b093e91d763ba99f8680457618fd); now we have a redirect to login right after we receive 401 code from the backedn;

> every sv2-ui restart logs everyone out (in-memory sessions). Fine for a laptop, but on Umbrel this happens on every update/reboot and the dashboard is often left open on a screen. I'm ok with it if we're explicit about it in the release notes, but maybe a sliding session would help too

I don't recommend increasing the complexity of the sessions at the moment, also consider that updates and reboots do  not happen very often on the Umbrel machine and our our app, so I think the tradeoff is acceptable. I've update the readme.(amended)

> no way to change the password while logged in

addressed in a new commit [7108423](https://github.com/stratum-mining/sv2-ui/pull/248/commits/71084235175a8b18f180ee4d0b7e2a3784aadf54)

> I'd add the usual show/hide toggle on the password fields

added in a new commit [3b2d3a6](https://github.com/stratum-mining/sv2-ui/pull/248/commits/3b2d3a6dc4050b3fe99f767d05c07387df8f1caf)

> Blocker: auth can be bypassed by changing the case of the path

> Blocker: auth can be bypassed by changing the case of the path.

that was a good one :-) I laughed. I assumed the framework would at least return a 404, but apparently case-sensitive routing has to be explicitly enabled and is not the default. That is pretty sad, especially for something security-related. see: https://expressjs.com/en/4x/api/application/#appset

Addressed in a new commit: https://github.com/stratum-mining/sv2-ui/pull/248/commits/b2d7edf76e43cafbbef3e3a80ba442930c763f49

And even with it enabled, Express still doesn't return a 404. It serves the SPA page instead. but is not leaking the data anymore;

Rest of the things are overkill for the scenarios. And this PR is getting unnecesseraly complex

GitGab19 (approved):

